PR #55, opened 2026-04-02
Support composite rules

### LINKED ISSUE #23: Support data type rules
We need something similar to Xtext's/Langium's data type rules. These should be implemented as specialized AST nodes that can be converted to a string representation.

msujew:
Closes https://github.com/TypeFox/fastbelt/issues/23

This is an adaption of data type rules from Xtext & Langium. We already had the idea in Langium to make data type rules a bit more explicit compared to Xtext - that idea never really went anywhere.

Instead, we now have the opportunity to structure these data type rules correctly:
1. Similarly to token based field, the value of a composite rule is always string-typed. The framework will not perform any conversion for adopters. Therefore, instead of calling them *data-type rules*, they are now simply *composite rules*.
2. Adds a dedicated AST node for composite rules. In the same vein, this PR slightly adjusts the type of the `Token` to support a common `StringUnit` interface that can be used by both `Token` and `CompositeNode` instances.
3. Properties that don't have a simple string as their value, but a `CompositeNode` use the `composite` type instead of `string` (I'm still on the fence about this - but we need some signifier for this).

I've used the statemachine language to demonstrate this a bit. I'll revert those changes (and maybe add a domainmodel example instead?) before merging this.

Also updates the lsp library to `v0.0.4` to support the `[]DefinitionLink` return type.


spoenemann:
What's the advantage of having a String node in the AST over a plain string value?

internal/grammar/grammar.fb
```diff
@@ -56,17 +57,20 @@ ReferenceType: "*" Type=[Interface:ID];
 
 SimpleType: Type=[Interface:ID];
 
-PrimitiveType: Type=("string" | "bool");
+PrimitiveType: Type=("string" | "bool" | "String");
```
Lotes: Why string is used twice?

ast.go
```diff
@@ -273,10 +276,67 @@ func AssignContainers(doc *Document, root AstNode) {
...
+func (node *StringNodeBase) IsStringNode() {}
+
+func (node *StringNodeBase) String() string {
+	if node.cache != "" {
+		return node.cache
+	}
+	// Construct the string value by concatenating the text of all tokens of the node
+	// Only need to do this once, as the tokens are usually not modified after parsing
+	var sb strings.Builder
+	for _, token := range node.Tokens() {
```
Lotes, line 347: I guess Whitespaces/Comments are already filtered out?
Is this meant to create the full name of a FQN rule?

internal/grammar/grammar.fb
Lotes: I guess the one is a string literal and the other is a string node.
Confusing. Take a different name: string_node, tokens, token_sequence, composite_string.

internal/grammar/grammar.fb
```diff
@@ -137,11 +141,22 @@ interface Action extends Element {
...
+StringGroup returns Element: StringElement ({Group.Elements+=current} Elements+=StringElement+)?;
+
+StringElement returns Element: (Keyword | RuleCall | "(" StringAlternatives ")") Cardinality=("*" | "+" | "?")?;
+
 // Move comments in front of RegexLiteral
 comment token SL_COMMENT: /\/\/[^\r\n]*/;
 comment token ML_COMMENT: /\/\*[^*]*\*\//;
 
-token String: /"[^"]+"/;
+token StringToken: /"[^"]+"/;
```
Lotes: Naming: Is this not a literal? Instead of Token?

examples/statemachine/statemachine.fb
```diff
@@ -52,10 +52,12 @@ interface Transition {
 
 Transition returns Transition:
     Event=[Event:ID] "=>"
-    State=[State:ID];
+    State=[State:FQN];
 
 token ID: /[_a-zA-Z][\w_]*/;
 
+string FQN: ID ("." ID)*;
```
Lotes: Can "String rules" be defined in a nested way or with fragments?

```
string FQN: ID ("." ID)*;
string FQNDeluxe: FQN "_" INT;
```

In my eyes they are standalone. Only tokens and keywords can be used.

internal/grammar/grammar.fb
Lotes: `composite token FQN: ID ("." ID)*;`

internal/generator/parser_generator.go
```diff
@@ -366,11 +388,21 @@ func generateRuleCallParser(node generator.Node, context *ParserGeneratorContext
...
+		switch t := target.(type) {
+		case grammar.Token:
+			n.AppendLine("token ", eq, " p.state.Consume(", GeneratedTokenIdxName(t), ")")
 			n.AppendLine("core.AssignToken(current, token, ", context.accessNames[ruleCall], ")")
-		} else if rule != nil {
-			n.AppendLine("result ", eq, " p.Parse", rule.Name(), "()")
+		case grammar.ParserRule:
+			n.AppendLine("result ", eq, " p.Parse", t.Name(), "()")
+		case grammar.StringRule:
+			if context.stringRule {
```
Lotes: Actually we know that we are in a grammar.StringRule. Why do we need that extra flag? Looks a bit suspicious.

Also the flag could be more precise in its name... it is IS-stringRule or WITHIN-stringRule? Or what else?

Lotes (commented):
I am not completely through yet.

I am wondering why we need this? Is it only for FQN in conjunction with cross-references? I seem to have some knowledge gap here...

I also have to think of users that put extra whitespace/comments within the FQN. The other option would be a chain of member calls.

examples/statemachine/statemachine.fb
msujew: You can nest them 👍

internal/generator/parser_generator.go
msujew: The difference is: Is only the target of the rule call a string rule, or are we also already within a string rule call? This requires a flag, since we cannot know this via the grammar hierarchy.

internal/grammar/grammar.fb
msujew: I like `composite`. I'll use that naming instead 👍

internal/generator/parser_generator.go
Lotes: Ok, I got it. Then, the name of the flag could be more meaningful: `withinStringRuleDecl`, `isGeneratingStringRule`

ast.go
msujew: Yes, we use the same logic in Langium/Xtext.

examples/statemachine/benchmark_test.go
```diff
@@ -156,5 +156,7 @@ func generateStatemachineContent(index int) (string, int) {
 		elementCount++
 	}
 
+	// TODO: remove this
```
spoenemann: Is this a leftover from some benchmark adaptation?

ast.go
```diff
@@ -273,10 +276,73 @@ func AssignContainers(doc *Document, root AstNode) {
...
+func (node *CompositeNodeBase) String() string {
+	// Fast path for inlining
+	if node.cache != "" {
+		return node.cache
+	}
+	// Slow path outlines the logic
+	return node.stringSlow()
+}
+
+func (node *CompositeNodeBase) stringSlow() string {
```
spoenemann, line 343: Multiple routines evaluating the String value in parallel might lead to multiple computations of the concatenated string. Question: is it worth synchronizing this with a mutex, or is the sync overhead too large in practice?

ast.go
```diff
@@ -273,10 +276,73 @@ func AssignContainers(doc *Document, root AstNode) {
...
+	// Construct the string value by concatenating the text of all tokens of the node
+	// Only need to do this once, as the tokens are usually not modified after parsing
+	var sb strings.Builder
+	for _, token := range node.Tokens() {
```
spoenemann, line 347: Are hidden tokens and comments already excluded from this slice?

spoenemann (approved):

examples/statemachine/benchmark_test.go
msujew: Yes, I've now removed the statemachine grammar changes again - and this todo is gone as well now.

ast.go
msujew: I've used `atomic.Pointer` instead here, which should solve the data race. In the worst case, the value is computed multiple times, but that's preferable to the overhead of a mutex here.

ast.go
msujew: Yes, `Tokens()` only contains the tokens that were actually used in the parsing process of the respective node.

PR #60, opened 2026-04-09
Add basic language testing support

### LINKED ISSUE #39: Add testing infrastructure
Currently, we mostly perform unit testing of individual isolated parts of the framework. To test more integrated features (linking, validation, LSP services), we need a better testing infrastructure similar to what we already have in Langium.

msujew:
Closes https://github.com/TypeFox/fastbelt/issues/39

Adds a basic testing infrastructure for languages. Informed by Langium and other projects. Currently, only supports testing of parsing, AST nodes, linking and diagnostics/validation. Other testing features (for example, LSP stuff) can come later.

Adds test files for the statemachine and grammar language.


examples/statemachine/statemachine_test.go
```diff
@@ -0,0 +1,205 @@
...
+	assert.Len(t, sm.Commands(), 2)
+	assert.Len(t, sm.States(), 2)
+}
+
+func TestFindAll(t *testing.T) {
+	f := fbtest.New(t, CreateServices())
+	doc := f.Parse(lightSwitch).AssertNoErrors()
+
+	events := fbtest.FindAll[Event](doc)
+	require.Len(t, events, 1)
```
Lotes, line 63: What is the difference between require and assert?

examples/statemachine/statemachine_test.go
```diff
@@ -0,0 +1,205 @@
...
+	assert.Equal(t, "flick", events[0].Name())
+
+	states := fbtest.FindAll[State](doc)
+	require.Len(t, states, 2)
+}
+
+func TestFindNamedNode(t *testing.T) {
+	f := fbtest.New(t, CreateServices())
```
Lotes: I like the grouping of lines.

For `examples/` code: you could prepend "arrange", "act", "assert" as a little guidance for adopters.

But I do not insist on it, just a nice-to-have...

internal/grammar/grammar_diagnostic_test.go
```diff
@@ -0,0 +1,238 @@
...
+// --- Rule and interface uniqueness ---
+
+func TestDuplicateRuleNames(t *testing.T) {
+	f := fbtest.New(t, CreateServices())
+	f.Parse(`
+		grammar Test;
+		interface Foo { Name string }
+		Foo: Name=ID;
+		Foo: Name=ID;
+	`+commonTokens).AssertDiagnostic(core.SeverityError, "A rule's name has to be unique. 'Foo' is used multiple times.")
```
Lotes: This chained call merges the ACT and ASSERT parts. Would like to see them separated.

internal/grammar/grammar_diagnostic_test.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+func TestDuplicateInterfaceNames(t *testing.T) {
+	f := fbtest.New(t, CreateServices())
+	f.Parse(`
+		grammar Test;
+		interface Foo { Name string }
+		interface Foo { Other string }
+	`+commonTokens).AssertDiagnostic(core.SeverityError, "An interface name has to be unique. 'Foo' is used multiple times.")
```
Lotes: Two more advanced ideas:
1. Distinguish the diagnostics by a code. This could create a habit to make a catalog of diagnostic codes. Then you just need to check for the code and for the arguments passed to the code. This could also make the following idea easier.
2. Would be cool to allow translations to other natural languages.

On the other hand: I do not think much people translate their diagnostic codes.
Having a catalog could be also restricting to the development process (slowing down).

Lotes (commented):
Nice idea with the range and index markers. Here are some questions, suggestions and ideas.

examples/statemachine/statemachine_test.go
msujew: `require` will fail the test and immediately leave the testing function. `assert` will fail, but continue with the testing function (and find potentially more issues in the code-under-test).

internal/grammar/grammar_diagnostic_test.go
msujew: I've actually refactored the whole diagnostics API. Let me know what you think!

Lotes: Thanks, adding the "code" avoids replicating the error message.

In order to make it translatable, we might need to add a map/dictionary with parameters (name + gap filler) and an exchangeable text with gaps using the keys from the map. What do you think about this?

msujew: Makes sense, but I don't think we need to worry about this right now. Most adopters don't test their diagnostics with multiple languages - we don't have such an infrastructure in Langium either.

examples/statemachine/statemachine_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+package statemachine
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	core "typefox.dev/fastbelt"
+	fbtest "typefox.dev/fastbelt/testing"
```
spoenemann: Is this an indicator that `fbtest` would be a better package name than `testing`?

BTW in the example languages, we should keep the package name `fastbelt` rather than renaming it to `core` – that renaming makes sense for internal framework dependencies.

examples/statemachine/statemachine_test.go
```diff
@@ -0,0 +1,205 @@
...
+  actions { turnOn }
+  flick => off
+end
+`
+
+func TestParsing(t *testing.T) {
+	f := fbtest.New(t, CreateServices())
+	f.Parse(lightSwitch).
+		AssertNoErrors().
+		AssertState(core.DocStateLinked)
```
spoenemann: I think this fluent API doesn't improve much on the readability. It would be more idiomatic to have every assertion in its own statement (just repeat the receiver of the assertion method).

examples/statemachine/statemachine_test.go
msujew: I've simply renamed it to `test` now. Didn't like `fbtest` really, since it would be `.../fastbelt/fbtest`.

spoenemann: Ok. However, `test` might look like the package contains actual tests. Maybe `testutil`?

`testing` would work, but collides with the stdlib package [testing](https://pkg.go.dev/testing).

We could also consider moving it into the `utils` folder. That doesn't change the user-facing package name, but may help clarifying its role.

spoenemann: Actually, seeing code usage like `test.New(...)`, that makes a lot of sense. I suggest keeping the name `test`, but inside `utils`. If we want something with a tighter scope, we could name it `langtest`.

internal/grammar/linking_test.go
```diff
@@ -0,0 +1,357 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+package grammar
```
spoenemann, line 5: We're in the `grammar` package, so the `grammar_` prefix shouldn't be necessary.

test/fixture.go
```diff
@@ -0,0 +1,290 @@
...
+// Content strings may embed position markers that are stripped before parsing.
+// Positions are recorded relative to the cleaned text.
+//
+//   - Range marker:  <|label:text|>  — spans "text"; label identifies the range.
+//   - Range shorthand: <|label|>     — equivalent to <|label:label|>.
+//   - Empty range:   <|label:|>      — zero-length span at the marker position.
+//   - Index marker:  <|label>        — single position (no text content).
+//
+// Range markers are tried before index markers when both share the same opening
+// delimiter (the default). Marker delimiters are configurable via [TestMarking].
```
spoenemann: Should we move package documentation to its own file `doc.go` so it's easier to find in the code base?

examples/statemachine/statemachine_test.go
msujew: Moving it to `utils` feels like it undermines that this is an actual feature of the framework. The code under `utils` so far has been used for "stuff that is useful for the framework internals, but can also be used by adopters", whereas `test` is specifically targeted at adopters.

I tried out `langtest`, but it doesn't read well in the tests IMO.

spoenemann (approved):

PR #62, opened 2026-04-13
Optimize linking using buckets

msujew:
Improves our linking infrastructure (through the board, i.e. both local symbols and exported/imported symbols) by using a generated type-optimized `SymbolContainer`. It stores types (only those actually required, not all!) in buckets, which are then merged on demand.

Since everything in the linking system has become `Symbol...` or `...Symbols`, I've chosen to rename `AstNodeDescription` to `SymbolDescription`. I believe this makes all of this a bit more coherent.

Performance improvements are clearly visible when looking at our statemachine benchmark. The overall build performance has increase by roughly ~20%. Linking by itself is now faster by 30%.


generated/services.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (b *GeneratedSrvContBlock) Generated() *GeneratedSrv {
+	return &b.generated
+}
+
+// GeneratedSrv contains the generated services for a specific language.
+type GeneratedSrv struct {
+	Lexer            lexer.Lexer
+	Parser           parser.Parser
+	SymbolContainers core.SymbolContainers
```
spoenemann, line 31: Shouldn't this be just a function instead of an interface with a method?

Comparison:
```go
srv.Generated().SymbolContainers.New()
```
vs.
```go
srv.Generated().NewSymbolContainer()
```

symbols.go
```diff
@@ -0,0 +1,110 @@
...
+// Language specific implementations optimize for this by indexing descriptions by the type of their AST node.
+type SymbolContainer interface {
+	// Attempts to put the given description into the container.
+	// Returns true if the description was added.
+	// The container is allowed to reject descriptions that it does not want to hold, for example because they are of the wrong type.
+	Put(desc *SymbolDescription) bool
+	// Returns an iterator over all descriptions in the container.
+	All() SymbolSeq
+	// Returns an iterator over all descriptions in the container that have a node of the given type.
+	Type(targetType reflect.Type) SymbolSeq
```
spoenemann: With this method name, I'd expect Type as return type. Maybe `ForType`?

spoenemann (commented):
Great improvement!

generated/services.go
msujew: I do like the interface more - but is just a preference. I feel like the system is a bit more coherent, if everything has a dedicated interface and we don't mix interfaces and factory function types within the service containers.

examples/statemachine/linker_gen.go
```diff
@@ -109,3 +112,58 @@ func (g *DefaultStatemachineModelReferencesConstructor) TransitionState(owner co
...
+		slices.Values(sc.Commands),
+	)
+}
+
+var TypeFor_State = reflect.TypeFor[State]()
+var TypeFor_Event = reflect.TypeFor[Event]()
+var TypeFor_Command = reflect.TypeFor[Command]()
+
+func (sc *StatemachineModelSymbolContainer) ForType(t reflect.Type) core.SymbolSeq {
+	switch t {
```
spoenemann, line 160: Does the type switch also work for subtypes (e.g. an interface embedding the super-interface)?

spoenemann (approved):

examples/statemachine/linker_gen.go
msujew: Yes, that's what the topological sorting is for. It's a bit cumbersome, but you can test it out by changing the grammar.

PR #64, opened 2026-04-25
Improve lexer performance using struct-array

msujew:
Performs some minor and a major performance optimizations:
1. (major) The `TokenSlice` uses the `Token` struct directly instead of using a pointer. This improves lexing performance by ~40%
2. (minor) Adds a running tokens-per-byte average to the lexer. This ensures (1) that we rarely try to increase the slice capacity and (2) keep the capacity as low as possible. Improves performance by ~3%.
3. (minor) changes the lookup from `slices.IndexFunc` to a custom loop. Does not influence performance - it seems like the Go compiler was able to inline the function call. However, this still makes everything easier to read.
4. (minor) Do not put the utf8 conversion behind a conditional. Improves performance by 3%.

### DIFF
--- a/document.go
+++ b/document.go
@@ -53,6 +53,7 @@ func NewDocument(textDoc textdoc.Handle) *Document {
 		ExportedSymbols: nil,
 		Data:            sync.Map{},
 		Tokens:          TokenSlice{},
+		Comments:        TokenSlice{},
 		ParserErrors:    []*ParserError{},
 		LexerErrors:     []*LexerError{},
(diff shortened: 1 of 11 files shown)


spoenemann (approved):
Great!

Do we need to update https://www.typefox.io/blog/fastbelt-introduction/ again 😁?

msujew:
> Do we need to update https://www.typefox.io/blog/fastbelt-introduction/ again 😁?

We could, though it's only a roughly ~4% performance increase in total.

PR #65, opened 2026-04-29
ATN construction

Lotes:
There is a big Compile Time ATN for construction and a slim Runtime ATN.

I also added a Markdown generator to be able to debug the ATN visually.


internal/grammar/atn.md
```diff
@@ -0,0 +1,781 @@
+# Runtime ATN for grammar
+
+## Grammar
```
Lotes, line 3: I think the Grammar rule misses a loopback, because we have a "*" as cardinality.

cmd/fastbelt/main.go
```diff
@@ -161,6 +161,14 @@ func runLegacyGenerate(args []string) error {
...
 		return err
 	}
+	if err := writeFile("atn", filepath.Join(outputPath, "atn_gen.go"),
+		generator.GenerateATN(grammar, packageName)); err != nil {
+		return err
+	}
+	if err := writeFile("atn-md", filepath.Join(outputPath, "atn.md"),
+		generator.GenerateATNMarkdown(grammar, packageName)); err != nil {
+		return err
+	}
```
msujew: Suggestion: I'd hide this behind a flag.

parser/allstar/grammar.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+// Production is the interface for all grammar elements.
+type Production interface {
+	Kind() ProductionKind
+	// Children returns the sub-productions, or nil for leaves.
+	Children() []Production
+	// Occurrence returns the 1-based index distinguishing multiple uses of the
+	// same production kind within one rule (0 for Alternative which is not indexed).
+	Occurrence() int
```
msujew: Question: Chevrotain needs this because of the way it works internally - however, I don't think we do. Instead, we can give all production a name instead (or at least a globally unique index).

parser/allstar/atn.go
msujew, line 1: Suggestion: Since we need all of this code in other parts of the parser (not just the lookahead), it might make sense to move it to the `parser` parent directory.

parser/atn_runtime.go
```diff
@@ -0,0 +1,66 @@
...
+	GetTarget() *RuntimeATNState
+	IsEpsilon() bool
+}
+
+// RuntimeAtomTransition fires on a specific token type.
+// Target, TokenTypeID, and CategoryMatches are exported so that generated
+// Go code in external packages can construct them as struct literals.
+type RuntimeAtomTransition struct {
+	Target          *RuntimeATNState
+	TokenTypeID     int
```
msujew: Suggestion: I don't think it's enough to just pass the target token type ID here, because we eventually want to implement token categories like in Chevrotain - see https://github.com/TypeFox/fastbelt/issues/49. For this, we need an actual reference to the token type.

parser/allstar/grammar.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Terminal consumes a single token.
+type Terminal struct {
+	TokenTypeID int
+	TokenName   string
+	Idx         int
+	// CategoryMatches holds token type IDs matched via category inheritance.
+	CategoryMatches []int
+}
```
msujew: Same here - just reference the token type.

parser/allstar/atn.go
```diff
@@ -0,0 +1,387 @@
...
+
+package allstar
+
+import "fmt"
+
+// ATNStateType is the discriminator for ATN state kinds.
+type ATNStateType int
+
+const (
+	ATNInvalidType    ATNStateType = 0
```
msujew: Suggestion: Use `iota`.

parser/allstar/grammar.go
```diff
@@ -0,0 +1,132 @@
...
+
+const (
+	ProdTerminal            ProductionKind = iota
+	ProdNonTerminal         ProductionKind = iota
+	ProdAlternative         ProductionKind = iota // one branch inside an Alternation; no occurrence index
+	ProdAlternation         ProductionKind = iota // OR(...)
+	ProdOption              ProductionKind = iota // OPTION(...)
+	ProdRepetition          ProductionKind = iota // MANY(...)
+	ProdRepetitionMandatory ProductionKind = iota // AT_LEAST_ONE(...)
+)
```
msujew:
```suggestion
const (
	ProdTerminal            ProductionKind = iota
	ProdNonTerminal         ProductionKind
	ProdAlternative         ProductionKind // one branch inside an Alternation; no occurrence index
	ProdAlternation         ProductionKind // OR(...)
	ProdOption              ProductionKind // OPTION(...)
	ProdRepetition          ProductionKind // MANY(...)
	ProdRepetitionMandatory ProductionKind // AT_LEAST_ONE(...)
)
```

parser/allstar/atn.go
```diff
@@ -0,0 +1,387 @@
...
+	// Populated for BlockEndState.
+	Start *ATNState
+
+	// Populated for RuleStartState.
+	Stop *ATNState
+}
+
+// ATN is the Augmented Transition Network built from a set of parser rules.
+type ATN struct {
+	DecisionMap      map[string]*ATNState
```
msujew: Suggestion: Can we make this an `[]*ATNState`? The parser and ATN should be able to simply index each decision. No need for a map lookup (this wasn't possible in Chevrotain, which is why it is a string there).

parser/allstar/grammar.go
```diff
@@ -0,0 +1,132 @@
+// Copyright 2025 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+// Package allstar implements the ALL(*) adaptive lookahead algorithm.
```
msujew: Minor: The `A` in `ALL` already stands for "adaptive".

internal/generator/atn_md_emit.go
```diff
@@ -0,0 +1,97 @@
...
+package allstar
+
+import (
+	"fmt"
+	"strings"
+
+	"typefox.dev/fastbelt/generator"
+)
+
+func EmitMarkdownSource(pkgName string, rtn *RuntimeATN, tokenTypeNames map[int]string) generator.Node {
```
msujew: Suggestion: This part of the code should live in the `internal` part of the framework - adopters shouldn't be able to call this function IMO.

internal/generator/atn_emit.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+// EmitGoSource returns Go source code for a function that constructs an
+// equivalent RuntimeATN without calling CreateATN.  The generated function
+// can be compiled into any package that imports the allstar package.
+//
+//	pkgName  – Go package name for the generated file (e.g. "myparser")
+//	funcName – name of the generated constructor function
+//	importPath – import path of the allstar package (e.g. "typefox.dev/fastbelt/parser/allstar")
+func EmitGoSource(pkgName, funcName, importPath string, rtn *RuntimeATN) generator.Node {
```
msujew: Suggestion: This part of the code should live in the `internal` part of the framework - adopters shouldn't be able to call this function IMO.

internal/generator/atn_emit.go
```diff
@@ -0,0 +1,180 @@
+// Copyright 2025 TypeFox GmbH
```
msujew: Please update the copyright headers to the current year.

parser/allstar/atn.go
Lotes: FTM a lot of stuff is inside the `generator` package. Only the Runtime ATN is in the `parser` package.

parser/allstar/grammar.go
Lotes: Removed

parser/allstar/grammar.go
Lotes: Removed

parser/atn_runtime.go
Lotes: FTM I removed the `CategoryMatches`. Eventually, we can link token types here, if needed.

internal/grammar/atn_gen.go
```diff
@@ -0,0 +1,2730 @@
+// Code generated by fastbelt. DO NOT EDIT.
+
+package grammar
+
+import "typefox.dev/fastbelt/parser"
+
+func BuildATN() *parser.RuntimeATN {
```
Lotes, line 40: Some randomness is still given with each generation, a deterministic solution would be hot.

internal/generator/atn_types.go
```diff
@@ -0,0 +1,91 @@
...
+func (t *RuleTransition) IsEpsilon() bool   { return true }
+
+// ATNHandle is an internal pair of (entry, exit) ATN states for a sub-network.
+type ATNHandle struct {
+	Left  *ATNState
+	Right *ATNState
+}
+
+// TokenType carries the type ID and category-match IDs for a token type.
+type TokenType struct {
```
Lotes: Need to replace this one...

parser/allstar/atn.go
Lotes: Good point, I will try to get rid of this. I would like to introduce a generated enumeration for this. Let's see...

internal/grammar/atn_gen.go
```diff
@@ -0,0 +1,2762 @@
...
+
+import "typefox.dev/fastbelt/parser"
+
+type ATNDecisionId int
+
+const (
+	Action_Alternatives_1                ATNDecisionId = 0
+	Action_Group_2                       ATNDecisionId = 1
+	Alternatives_Group_2                 ATNDecisionId = 2
+	Alternatives_Group_3                 ATNDecisionId = 3
```
Lotes: The counter inside the name looks broken...

Lotes: The counter refers to the index of all parts in a rule. Only a few are used for the decision map.

internal/atn/builder.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	atn.DecisionMap[lookaheadName] = start
+
+	return &ATNHandle{Left: start, Right: end}
+}
+
+func (rb *ATNRuleBuilderImpl) MakeBlock(alts []*ATNHandle) *ATNHandle {
+	for i := 0; i < len(alts)-1; i++ {
+		handle := alts[i]
```
Lotes, line 194: @msujew Not very sure about the [original implementation](https://github.com/TypeFox/chevrotain-allstar/blob/cf376d1f28860ddd183cd3225cf5dee223f549ea/src/atn.ts#L513)...
May need to redo it...

internal/atn/atn_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func TestStarRule(t *testing.T) {
+	atn, rules, tokenTypes := FixtureATN(t, GrammarTemplate(`
+		interface Start { Property []Rule }
+		interface Rule { Name string }
+		Start: Property+=Rule*;
+		Rule: Name=ID;
+	`))
+	RequireATNRecognizes(t, atn, rules, tokenTypes, "Start", []string{}, 1)
```
Lotes, line 50: Add tests with 2 or more paths…

internal/generator/atn.go
```diff
@@ -0,0 +1,300 @@
...
+	rb ATNRuleBuilder,
+	cr grammar.CrossRef,
+	names map[grammar.Element]string,
+	tokenTypes map[string]TokenType,
+) (*ATNHandle, error) {
+	// Use the explicitly named rule if present, otherwise fall back to "ID".
+	name := "ID"
+	if cr.Rule() != nil && cr.Rule().Rule() != nil && cr.Rule().Rule().Text() != "" {
+		name = cr.Rule().Rule().Text()
+	}
```
msujew: Suggestion: Fastbelt does not allow optional cross ref terminals

internal/generator/atn.go
```diff
@@ -0,0 +1,300 @@
...
+	tokenTypes map[string]TokenType,
+	rulesByName map[string]*grammar.ParserRule,
+) (*ATNHandle, error) {
+	if rc.Rule() == nil {
+		return nil, fmt.Errorf("RuleCall has no rule reference")
+	}
+	name := rc.Rule().Text()
+	if name == "" {
+		return nil, fmt.Errorf("RuleCall rule reference has empty text")
+	}
```
msujew: Suggestion: We could propagate errors through the system, but we generally can only arrive here after the grammar validation has passed - i.e. we can expect the grammar/input to be syntactically and semantically valid. Doesn't it make more sense to simply panic in case of errors? This would cut down on complexity a lot in this file.

internal/grammar/atn_gen.go
```diff
@@ -0,0 +1,2762 @@
+// Code generated by fastbelt. DO NOT EDIT.
+
+package grammar
+
+import "typefox.dev/fastbelt/parser"
+
+type ATNDecisionId int
+
+const (
+	Action_Alternatives_1                ATNDecisionId = 0
```
msujew: Suggestion: `iota`

parser/atn_runtime.go
```diff
@@ -0,0 +1,83 @@
...
+	ATNPlusLoopBack
+	ATNLoopEnd
+)
+
+// RuntimeATNState holds only the fields required for prediction at runtime.
+// Back-pointers to the build-time grammar objects (Rule, Production) and
+// structural scaffolding (End, Loopback, Start, Stop) are absent.
+//
+// RuleName and Prod* are populated only on decision states and are used
+// exclusively for ambiguity-report messages.
```
msujew: Question: This comment does not seem to be correct? They are never populated in the code as of right now.

internal/atn/fixture.go
```diff
@@ -0,0 +1,136 @@
...
+func GrammarTemplate(rules string) string {
+	return "grammar Test;\n" + rules + `
+token ID: /[a-zA-Z_][a-zA-Z0-9_]*/;
+token NUMBER: /[0-9]+/;
+hidden token WS: /[ \n\r\t]+/;
+`
+}
+
+func FixtureATN(t *testing.T, grammarStr string) (*ATN, map[string]*grammar.ParserRule, map[string]TokenType) {
+	f := test.New(t, grammar.CreateServices())
```
msujew, line 38: Suggestion: Mark these as `t.Helper()` functions.

msujew (commented):
Looking really promising already, just a few comments, see below 👍

internal/atn/builder.go
msujew: IIRC the code in the original implementation is supposed to simplify stuff like this:

image

In this case, all states 254-261 are not required for properly evaluating the ATN. We can go straight from 262 to 263 using the rule transitions.

internal/grammar/atn_gen.go
```diff
@@ -0,0 +1,2762 @@
...
+	}
+	states[46].Transitions = []parser.RuntimeTransition{
+		&parser.RuntimeEpsilonTransition{Target: states[285]},
+	}
+	states[48].Transitions = []parser.RuntimeTransition{
+		&parser.RuntimeEpsilonTransition{Target: states[301]},
+	}
+	states[50].Transitions = []parser.RuntimeTransition{
+		&parser.RuntimeAtomTransition{Target: states[51], TokenTypeID: 20},
+	}
```
msujew, line 1609: Suggestion: I meant this with [my comment](https://github.com/TypeFox/fastbelt/pull/65#discussion_r3167502989). That way we don't need to embed any duplicated token information into the transition. We can really just point to the token type pointer.
```suggestion
	states[50].Transitions = []parser.RuntimeTransition{
		&parser.RuntimeAtomTransition{Target: states[51], TokenType: Keyword_grammar},
	}
```

parser/atn_runtime.go
Lotes: I remove the ftm

internal/atn/builder.go
Lotes: MakeBlock is about concatenation, not alternatives.
Is this so much faster? I would prefer simple code.

Modelling middleware for the ATN could be a thing, that could be designed to optimize the ATN.

msujew: Right, and the idea is to skip 254-255, 256-257, etc. since they don't serve any value in the ATN. In the end, this makes the subset construction algorithm - and all other algorithm that need to traverse the ATN - slower. I would assume by quite a significant margin actually, with more gains the *deeper* the grammar becomes. 

IMO, we should do everything, even if complicated, to improve the performance of the runtime-code in favor of more complexity in the compile-time code.
And to be honest, the simplification code isn't that difficult, though feel free to add further comments here.

> Modelling middleware for the ATN could be a thing, that could be designed to optimize the ATN.

Sounds more complex than just adding the optimization in there actually ;)

Lotes: True. The middleware thingy sounds more complicated, when looking only at this single optimization. When we would add more of these optimizations the code becomes more unreadable. The middleware or middleend was an idea to introduce a optimization module.
But I remember that there is also a minimization strategy for DFAs. Maybe we can do something similar here?
Middleware or a big optimization step afterwards, so or so, I like this code here to be in a simple shape. But I could reintroduce the code with a explanation and a TODO, that we can refactor in a future PR.

msujew: I'd be a bit hesitant to change the structure of the ATN too much. After all, we need it later to perform code completion and error recovery.

Lotes: Then let‘s postpone the refactoring until we have the necessary features with tests.

msujew: Fine with me 👍

Lotes: Added the optimization as such.

internal/atn/builder.go
```diff
@@ -0,0 +1,293 @@
...
+			rb.NewEpsilonTransition(start, end)
+		}
+	}
+
+	atn.DecisionMap[lookaheadName] = start
+
+	return &ATNHandle{Left: start, Right: end}
+}
+
+func (rb *ATNRuleBuilderImpl) MakeBlock(alts []*ATNHandle) *ATNHandle {
```
Lotes: Rename to „concatenation“ or rename „concatenation“ to „block“…

internal/atn/fixture.go
```diff
@@ -0,0 +1,147 @@
...
+	t.Helper()
+	f := test.New(t, grammar.CreateServices())
+	doc := f.Parse(grammarStr).AssertNoErrors()
+	grammr := doc.Root().(grammar.Grammar)
+	atn, rules := CreateATN(grammr, TokenTypeIds)
+	node := EmitMarkdownSource("test", atn, TokenTypeNames)
+	content := node.String()
+	if err := os.WriteFile("atn.test.md", []byte(content), 0644); err != nil {
+		t.Fatalf("failed to write ATN markdown file: %v", err)
+	}
```
msujew, line 49: Question: This doesn't seem to be part of the automated test, right? Is this for manual inspection? Doesn't this override everything for each test?

internal/generator/lexer_generator.go
```diff
@@ -6,6 +6,7 @@ package generator
 
 import (
 	"fmt"
+	"maps"
```
msujew, line 9: Please also regenerate the statemachine language.

msujew (approved):
Looks good, just one question, see below.

internal/atn/fixture.go
Lotes: True. This is for exact this purpose. Debugging the current failing test.
Shall I improve something?

PR #66, opened 2026-04-30
Refactored service binding

spoenemann:
I feel the service container approach introduced in #6 is too complicated. I propose to simplify it as done in this PR.

We define a `service.Container` utility where you can register (`MustPut` or `MustOverride`) and retrieve (`Get` or `MustGet`) service instances. There's no "injection" – every service simply keeps a pointer to the service container and retrieves things when they're needed.

## Performance Impact

I added a helper program to run a standardized benchmark:

```
go run ./internal/bench -runs 30 -cpu 1
```

On main:
```
MB/s:        min 3.380, avg 3.465, max 3.520
ms/resource: min 1.4710, avg 1.4953, max 1.5310
```

On this PR:
```
MB/s:        min 3.400, avg 3.456, max 3.510
ms/resource: min 1.4740, avg 1.4991, max 1.5230
```

This is 0.25% slower than before, probably due to the hash map lookups.


examples/statemachine/parser_gen.go
```diff
@@ -5,26 +5,27 @@ package statemachine
...
 
 func (p *Parser) references() StatemachineModelReferencesConstructor {
-	return p.srv.StatemachineModelLinking().ReferencesConstructor
+	return service.MustGet[StatemachineModelReferencesConstructor](p.sc)
 }
 
 func (p *Parser) Parse(document *core.Document) *parser.ParseResult {
-	cp := &Parser{srv: p.srv, state: parser.NewParserState(document.Tokens)}
+	cp := &Parser{sc: p.sc, state: parser.NewParserState(document.Tokens)}
 	result := cp.ParseStatemachine()
```
msujew, line 20: Suggestion: I assume the minor performance degradation comes from repeated access to the `MustGet` when creating a reference. Can we store the `<language>ReferencesConstructor` directly on the parser when re-constructing it during `Parse`?

util/service/service.go
```diff
@@ -0,0 +1,98 @@
...
+		panic(err)
+	}
+	return service
+}
+
+// MustPut puts a service into the container.
+// It panics if the container is sealed or the service already exists.
+//
+// TODO make this a method once generic methods are supported in Go.
+func MustPut[T any](container *Container, service T) {
```
msujew: Suggestion: IMO `MustPut` only makes sense if we also have a `Put(...) error` function. Can't we just call this `Put`?

util/service/service.go
```diff
@@ -0,0 +1,98 @@
...
+		panic(fmt.Sprintf("service %s already exists", t.Name()))
+	}
+	container.services[t] = service
+}
+
+// MustOverride overrides a service in the container.
+// It panics if the container is sealed or the service does not exist.
+//
+// TODO make this a method once generic methods are supported in Go.
+func MustOverride[T any](container *Container, service T) {
```
msujew: Suggestion: Same, just `Override`?

workspace/document_manager.go
```diff
@@ -25,14 +26,19 @@ type DocumentManager interface {
 	All() iter.Seq[*core.Document]
 	// Deletes the document with the given URI and returns it, or nil if it did not exist.
 	Delete(uri core.URI) *core.Document
+	// Clears all documents from the manager.
+	Clear()
 }
 
+// DefaultDocumentManager is the default implementation of [DocumentManager].
 type DefaultDocumentManager struct {
+	sc        *service.Container
```
msujew, line 35: Question: Why?

util/service/service.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"fmt"
+	"reflect"
+)
+
+// Container holds a collection of services to be used in the framework.
+type Container struct {
+	services map[reflect.Type]any
+	sealed   bool
+}
```
msujew, line 18: Suggestion: I'd appreciate a `func (c *Container) All() iter.Seq[any]` for future use cases.

linking/linker.go
```diff
@@ -19,13 +20,17 @@ type Linker interface {
 	Link(ctx context.Context, document *core.Document)
 }
 
-type DefaultLinker struct{}
+// DefaultLinker is the default implementation of [Linker].
+// It resolves all references in the document.
+type DefaultLinker struct {
+	sc *service.Container
+}
```
msujew, line 27: Question: Why add the `sc` by default?

msujew (commented):
Thanks, makes this a lot easier to understand.

workspace/document_manager.go
spoenemann: It doesn't hurt to have the service container in every service, and it avoids breaking the API if we need it later.

linking/linker.go
spoenemann: Avoid breaking API later – we might need to add a dependency to another service.

spoenemann:
With the service caching, we're even 0.3% faster than on main:

```
MB/s:        min 3.430, avg 3.477, max 3.580
ms/resource: min 1.4460, avg 1.4902, max 1.5110
```

msujew (approved):
Makes sense that we're a tiny bit faster now 👍 

The new system looks much simpler - thank you!

PR #67, opened 2026-05-04
Implement rename provider

### LINKED ISSUE #27: Rename LSP feature
Based on top of https://github.com/TypeFox/fastbelt/issues/24.

msujew:
Closes https://github.com/TypeFox/fastbelt/issues/27

In addition to implementing the `RenameProvider` this PR adds a few other things:
1. Adds an abstraction service for finding names and references to be used by the other LSP services
2. Adds test functions for rename, definition and references requests
3. Adds tests to the statemachine language for these
4. Fixes some marker extraction logic errors.


examples/statemachine/statemachine_references_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	doc.AssertReferences("flick")
+}
+
+func TestReferencesOnReference(t *testing.T) {
+	f := test.New(t, CreateLspServices())
+	doc := f.Parse(`
+		statemachine Toggle
+
+		events <|flick|>
```
Lotes, line 39: I am a bit confused about the usage of ranges and indices here.
In the fourslash framework we used <|xxx|> where xxx was a declaration and all indices <|xxx> pointed to a reference of the xxx declaration.

"flick" is an event, but you are using <|...|> also at other positions that are not declarations.

server/references_provider.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 TypeFox GmbH
+// Copyright 2026 TypeFox GmbH
```
Lotes, line 1: Other PR: Add "fix/check header" scripts + workflow

ast_test.go
```diff
@@ -0,0 +1,94 @@
...
+	words := [16]string{
+		"alpha", "beta", "gamma", "delta",
+		"epsilon", "zeta", "eta", "theta",
+		"iota", "kappa", "lambda", "mu",
+		"nu", "xi", "omicron", "pi",
+	}
+
+	// Build the node pool using a deterministic seed so the benchmark is
+	// reproducible, but varied enough to defeat any trivial optimisation.
+	src := rand.New(rand.NewPCG(0xdeadbeef, 0xcafebabe))
```
Lotes, line 82: `0x1cec0ffee` xD

examples/statemachine/statemachine_references_test.go
msujew: Ranges don't have to be declarations, see for example the semantic token testing in the fourslash framework. In particular, in Fastbelt, it will simply use the start index of the range - eventually we can change it so every index covered by the range should do the expected work (though that's a bit difficult with the rename, since it has side effects).

server/name_finder.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type FoundName struct {
+	// The unit that contains the token that was used to start the search.
+	Source core.StringUnit
+	// The unit that contains the name of the symbol that was found.
+	Target core.StringUnit
+}
+
+// The [NameFinder] is responsible for finding the source and target [core.StringUnit] for a given token.
```
spoenemann: As far as I could find out, we shouldn't use a reference like `[NameFinder]` pointing at the thing being documented. But we should always start the doc comment with the element's name.
```suggestion
// NameFinder is responsible for finding the source and target [core.StringUnit] for a given token.
```

util/extiter/extiter.go
```diff
@@ -17,6 +17,10 @@ func Empty[T any]() iter.Seq[T] {
 	return func(yield func(T) bool) {}
 }
 
+func Of[T any](elements ...T) iter.Seq[T] {
+	return slices.Values(elements)
```
spoenemann, line 21: Why not use `slices.Values` directly?

examples/statemachine/services.go
```diff
@@ -30,3 +31,12 @@ func CreateServices() *service.Container {
 	sc.Seal()
 	return sc
 }
+
+// CreateServices creates a service container for the statemachine language to be used in the language server.
```
spoenemann:
```suggestion
// CreateLspServices creates a service container for the statemachine language to be used in the language server.
```

server/references_finder.go
```diff
@@ -0,0 +1,52 @@
...
+
+type DefaultReferencesFinder struct {
+	sc *service.Container
+}
+
+func NewDefaultReferencesFinder(sc *service.Container) ReferencesFinder {
+	return &DefaultReferencesFinder{sc: sc}
+}
+
+func (rp *DefaultReferencesFinder) Find(ctx context.Context, target core.AstNode, options FindReferencesOptions) iter.Seq[*core.ReferenceDescription] {
```
spoenemann: Receiver should be `rf` for ReferencesFinder, or just `s` for Service.

spoenemann (approved):
Great!

util/extiter/extiter.go
msujew: API convenience, `slices.Values` only accepts a slice, while `Of` will accepts atomic elements. I believe the compiler should inline this anyway, so it shouldn't create any runtime overhead.

PR #69, opened 2026-05-12
Add folding range lsp support

### LINKED ISSUE #26: Folding feature
See Langium's implementation.

ssmifi:
Closes https://github.com/TypeFox/fastbelt/issues/26
and implements the folding feature, which is basically a port from Langium. 

Customization can be done via the `FoldingRangeFilter` interface.


server/folding_range_provider.go
```diff
@@ -0,0 +1,146 @@
...
+	// IncludeLastFoldingLine determines whether the folding range should include its last line for AST nodes.
+	// Return false to exclude the last line (e.g., for nodes ending in closing braces).
+	IncludeLastFoldingLine(node core.AstNode) bool
+
+	// IncludeLastFoldingLineForComment determines whether comment folding ranges should include the last line.
+	// Return false to exclude the last line (typically desired for multi-line comments).
+	IncludeLastFoldingLineForComment() bool
+}
+
+type DefaultFoldingRangeFilter struct{}
```
msujew, line 37: Minor: Add a comment what the default implementation does.

server/folding_range_provider.go
```diff
@@ -0,0 +1,146 @@
...
+	if segment.Range.End.Line-segment.Range.Start.Line < 2 {
+		return nil
+	}
+
+	foldRange := segment.Range
+	if !includeLastLine {
+		foldRange = core.TextRange{
+			Start: segment.Range.Start,
+			End:   core.TextLocation{Line: segment.Range.End.Line - 1, Column: segment.Range.End.Column},
+		}
```
msujew: Suggestion: I don't think the logic is correct here. It should use the `Column` value at the end of the line.

server/folding_range_provider_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+		if *fr.StartLine == uint32(markerRange.Start.Line) && *fr.EndLine == uint32(markerRange.End.Line) {
+			if label == "comment" {
+				assert.Equal(t, "comment", fr.Kind, "Comment folding should have kind='comment'")
+			}
+			return
+		}
+	}
+	t.Errorf("Should have folding range for marker '%s' (lines %d-%d)", label, markerRange.Start.Line, markerRange.End.Line)
+}
```
msujew: Suggestion: I would like to see a `AssertFoldingRanges` in the `Doc` struct instead of extracting the markers directly. See [my other PR](https://github.com/TypeFox/fastbelt/pull/67) for inspiration.

msujew (commented):
Thanks, I like the filter strategy approach! A few suggestion, see below 👍

server/folding_range_provider.go
ssmifi: You are right, this is wrong. Thanks. I think we can just omit the end char. LSP should default to the whole line then.

server/folding_range_provider_test.go
ssmifi: Sounds good. Should I wait until you've merged the rename provider to make changes/merges later easier?

msujew: I don't think you need my changes. Merge conflicts should be rather easy to resolve. You can just go ahead :)

test/fixture.go
```diff
@@ -158,7 +158,7 @@ func (f *Fixture) ParseAll(uriContentPairs ...string) []*Doc {
 }
 
 func (f *Fixture) newDoc(doc *core.Document, ranges []RangeMarker, indices []IndexMarker) *Doc {
-	return &Doc{Document: doc, Ranges: ranges, Indices: indices, ctx: f.ctx, t: f.t}
+	return &Doc{Document: doc, Ranges: ranges, Indices: indices, ctx: f.ctx, t: f.t, fixture: f}
```
msujew, line 161: Suggestion: Why pass the `fixture` if all you need is `t` anyway (which already exists in `Doc`)?

test/doc_fixture.go
```diff
@@ -424,3 +426,38 @@ func MustFindReference[T core.AstNode](d *Doc, label string) *core.Reference[T]
 	}
 	return r
 }
+
+// AssertFoldingRanges verifies that the given folding ranges contain entries for all specified marker labels.
+// Returns the Doc for chaining.
+func (d *Doc) AssertFoldingRanges(result []lsp.FoldingRange, labels ...string) *Doc {
```
msujew: Suggestion: Instead of the test passing in the folding ranges, the method should retrieve them by itself. This keeps the tests much smaller. See as an example:

https://github.com/TypeFox/fastbelt/blob/ce62095d4bfec23a1da51df472eb72d40b54380f/test/doc_fixture.go#L104-L122

internal/grammar/folding_range_provider_test.go
msujew, line 1: Suggestion: Move this test file to `internal/grammar`. If we were to add references to `fastbelt/server` inside of `fastbelt/internal/grammar`, this test suite fails to compile due to cyclic dependencies.

test/doc_fixture.go
ssmifi: But then I can pass the fixture, yes? It's needed for the renameProvider anyways, right?

msujew: It's not required. But it's alright 👍

msujew (approved):
Thanks, works well 👍

PR #70, opened 2026-05-16
Implement parser error recovery

### LINKED ISSUE #80: Optional keyword leads to errors in generated parser
When changing the grammar definition to

```
Grammar: "grammar" Name=ID ";"?
```

with the added `?` after the semicolon, the generated parser has a compiler error.

msujew:
Closes https://github.com/TypeFox/fastbelt/issues/80

Does as the title says. The implementation is heavily inspired by ANTLR4's implementation, using the same mechanism for the most part. Implements additional caching/precomputation steps to improve performance:
1. On my machine, the parser only benchmark runs at 100MB/s.
2. Without caching, the parser benchmark runs at ~45MB/s.
3. With caching, we reach 91MB/s.

Supporting error recovery thus has a 10% performance penalty on parsing due to having to check the `followSet` before and during loops and rule calls. This is definitely within expectations and does not impede on our performance goals.

Performs minor changes to the ATN construction to better facillitate accessing the states in the parser generator. It should have no noticable impact on further ATN work (hopefully).

Can be tested quite well by just playing around with the language servers of the grammar and statemachine language. The PR also contains a few automated tests for the statemachine language.


parser/parser.go
```diff
@@ -106,3 +186,43 @@ func (p *ParserState) Lookahead(value LLkLookahead) int {
...
+// optional/loop guards.
+func (p *ParserState) Sync(decisionStateIdx int) {
+	p.recovery.Sync(p, decisionStateIdx)
+}
+
+// FollowSet returns the union of NextTokensAt for every frame on the follow-state stack.
+// The returned slice is indexed by TokenType.Id; out-of-range indices indicate
+// "not in the follow set".
+func (p *ParserState) FollowSet() []bool {
+	if p.atn == nil {
```
Lotes: In which scenario the ATN is nil? Why this check?

examples/statemachine/parser_gen.go
```diff
@@ -104,61 +106,73 @@ func (p *Parser) ParseStatemachine() Statemachine {
...
-		token := p.state.Consume(Token_ID_Idx)
+		token := p.state.Consume(Token_ID)
 		core.AssignToken(current, token, StatemachineInitID_0)
 		if token != nil {
-			current.SetInit(p.referencesConstructor().StatemachineInit(current, token))
+			current.SetInit(p.referencesConstructor.StatemachineInit(current, token))
 		}
 	}
 	{
+		p.state.Sync(26)
```
Lotes, line 167: Replace magic numbers with constants.

parser/atn_runtime.go
```diff
@@ -73,4 +77,101 @@ type RuntimeATN struct {
...
+	if stateIdx < 0 || stateIdx >= len(atn.nextTokensCache) {
+		return nil
+	}
+	return atn.nextTokensCache[stateIdx]
+}
+
+func (atn *RuntimeATN) buildNextTokensCache() {
+	maxId := 0
+	for _, st := range atn.States {
+		if st == nil {
```
Lotes: Is this really possible?

parser/error_messages.go
```diff
@@ -0,0 +1,56 @@
...
+// DefaultErrorMessageProvider produces English diagnostic messages. It is the
+// default ErrorMessageProvider used when none is supplied to NewParserState.
+type DefaultErrorMessageProvider struct{}
+
+func NewDefaultErrorMessageProvider() DefaultErrorMessageProvider {
+	return DefaultErrorMessageProvider{}
+}
+
+func (DefaultErrorMessageProvider) UnexpectedEndOfInput(expected *core.TokenType) string {
+	return "Unexpected end of input, expected '" + expected.Name + "'."
```
Lotes, line 39: Nice file for translations

parser/error_recovery.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	state.InError = false
+	state.ErrorRecoveryMode = false
+}
+
+// Attempts to recover from a parser error by calling the current stategy's Recover method.
+// Immediately returns if the recovery fails.
+// If not within in error, it tries to ensure that the upcoming token is valid for the current decision state.
+// This ensures that the parser can continue to make progress and doesn't get stuck on a bad token.
+func (DefaultErrorRecovery) Sync(state *ParserState, decisionStateIdx int) {
```
Lotes: Question: It might be confusing to use the term "parser STATE" here, because we also use it for ATN and DFA and NFA states. So something that is an automaton.

Maybe CONTEXT is an alternative (thesaurus search: environment, frame, surroundings, situation). This is more an opinion / observation. For a moment I was wondering why all ATN states have a recovery strategy. Then I saw that it is the state of the parser.

Lotes: Prefixing the variable could be also a solution: "parserState" instead of "state"

parser/error_recovery.go
```diff
@@ -0,0 +1,143 @@
...
+	if state.InError {
+		state.RecoveryStrategy().Recover(state)
+		if state.InError {
+			return
+		}
+	}
+	tok := state.LARaw(1)
+	valid := state.atn.NextTokensAt(decisionStateIdx)
```
Lotes: "valid" looks like a single boolean. Name it `validTokens`

parser/error_recovery.go
```diff
@@ -0,0 +1,143 @@
...
+	// This is the expected case: LA(1) is in the set of valid tokens.
+	// We can immediately return, the parser is valid and ready to continue.
+	// This is the hot path for error-free parsing, so it's important that this check is fast.
+	if tok == nil || tokenInSet(valid, tok.TypeId) {
+		return
+	}
+	follow := state.FollowSet()
```
Lotes: Suggestion: `followSetTokens`? Follow sets are always about token (types), or?
During review it is not easy to display the types. Changing the name to something type-reflecting could be a help.

Lotes (commented):
Not entirely through yet, this is just part I.

examples/statemachine/parser_gen.go
msujew: That's a good point, but I would prefer to do this in a separate PR.

spoenemann (approved):
Seems to work well. I'm still getting server crashes when editing the grammar, though.

Regarding error messages: "Missing A, got B" sounds a bit like caveman style 😅

msujew:
@spoenemann Those are fixed in https://github.com/TypeFox/fastbelt/pull/79 I believe.

PR #73, opened 2026-05-18
Fix ATN generation for composite rules

### LINKED ISSUE #72: ATN + composites = panic
hi, porting our langium grammar to fastbelt (we are using composites a lot)
then we face following panic in atn gen:

https://github.com/TypeFox/fastbelt/pull/71/changes

msujew:
Closes https://github.com/TypeFox/fastbelt/issues/72 and closes https://github.com/TypeFox/fastbelt/pull/71.

Simply adds support to composite types in the ATN generation by using the common `AbstractRuleWithBody` type.

Also removes a bunch of unnecessary interface-pointers.

### DIFF
--- a/internal/atn/atn.go
+++ b/internal/atn/atn.go
@@ -164,10 +164,7 @@ func convertRuleCall(
 	lookaheadName := rb.GetLookaheadNameByElement(rc)
 
 	switch typed := rule.(type) {
-	case grammar.CompositeRule:
-		handle := rb.RuleRef(typed)
-		return wrapWithCardinality(rb, handle, cardinality, lookaheadName), nil
-	case grammar.ParserRule:
+	case grammar.AbstractRuleWithBody:
 		handle := rb.RuleRef(typed)
 		return wrapWithCardinality(rb, handle, cardinality, lookaheadName), nil
 	case grammar.Token:
--- a/internal/atn/atn.go
+++ b/internal/atn/atn.go
@@ -5,6 +5,7 @@
 package atn
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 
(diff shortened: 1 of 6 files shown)


internal/atn/atn_test.go
```diff
@@ -149,3 +149,21 @@ func TestThreePaths(t *testing.T) {
...
+		composite SimpleComposite: ID;
+		Start: Name=SimpleComposite;
+	`, false)
+	RequireATNRecognizes(t, atn, rules, tokenTypes, "Start", []string{"ID"}, 1)
+}
+
+func TestCompositeRuleWithAlternativesRef(t *testing.T) {
+	atn, rules, tokenTypes := FixtureATN(t, `
+		interface Start { Name string }
+		composite IDOrNumber: ID | NUMBER;
```
Lotes, line 165: When I read "composite" I would expect a concatenation instead of a alternation as non-simple test. Can you add one?

internal/atn/builder.go
```diff
@@ -82,13 +82,13 @@ func (rb *ATNRuleBuilderImpl) GetLookaheadNameByElement(el grammar.Element) stri
 	panic("no lookahead name found for element")
 }
 
-func (b *ATNBuilderImpl) DeclareRule(rule *grammar.ParserRule) ATNRuleBuilder {
+func (b *ATNBuilderImpl) DeclareRule(rule grammar.AbstractRuleWithBody) ATNRuleBuilder {
```
Lotes, line 85: Thanks for removing the unnecessary pointers.

internal/atn/atn.go
```diff
@@ -152,27 +160,28 @@ func convertRuleCall(
...
 
-	if rule := rb.GetRuleByName(name); rule != nil {
-		handle := rb.RuleRef(rule)
+	switch typed := rule.(type) {
+	case grammar.CompositeRule:
+		handle := rb.RuleRef(typed)
+		return wrapWithCardinality(rb, handle, cardinality, lookaheadName), nil
+	case grammar.ParserRule:
+		handle := rb.RuleRef(typed)
 		return wrapWithCardinality(rb, handle, cardinality, lookaheadName), nil
```
Lotes, line 169: Can we unite those two cases, I do not see a difference.
Was is `case grammar.CompositeRule, grammar.ParserRule:`?

Lotes (commented):
The composite was somewhere wide back in my head. Thanks for adding it.

internal/atn/atn_test.go
msujew: Sure, I've added another test.

internal/atn/atn_test.go
```diff
@@ -167,3 +167,12 @@ func TestCompositeRuleWithAlternativesRef(t *testing.T) {
...
 	RequireATNRecognizes(t, atn, rules, tokenTypes, "Start", []string{"ID"}, 1)
 }
+
+func TestCompositeRuleWithMultipleTokens(t *testing.T) {
+	atn, rules, tokenTypes := FixtureATN(t, `
+		interface Start { Name composite }
+		composite FQN: ID ("." ID)*;
+		Start: Name=FQN;
+	`, false)
+	RequireATNRecognizes(t, atn, rules, tokenTypes, "Start", []string{"ID"}, 1)
```
Lotes: code +=

```
RequireATNRecognizes(t, atn, rules, tokenTypes, "Start", []string{
  "ID", "Dot", "ID", "Dot", "ID"
}, 1)
```

Please? :)

Lotes (approved):
Just add that longer parsing test assertion (or even some more :D). Looks good to me.

PR #75, opened 2026-05-20
Add document symbol feature

### LINKED ISSUE #25: Outline/Document Symbol feature
See Langium's implementation of the `textDocument/documentSymbols` request.

ssmifi:
Closes https://github.com/TypeFox/fastbelt/issues/25
and implements the document symbol provider, which is basically a port from Langium.

There is some customization possible and also implements the `nameProvider` and the `nodeKindProvider`. 
The document symbol provider loads these lazily allowing for service overrides after construction. But we can ofc discuss the architecture. :)


server/name_provider.go
```diff
@@ -0,0 +1,43 @@
...
+type NameProvider interface {
+	// GetName returns the name of the given AST node.
+	// Returns empty string if the node has no name.
+	GetName(node core.AstNode) string
+
+	// GetNameSegment returns the text segment for the name portion of the node.
+	// This is used for precise selection ranges in document symbols.
+	// Returns nil if the name segment cannot be determined.
+	GetNameSegment(node core.AstNode) *core.TextSegment
+}
```
msujew: Suggestion: We recently just removed the `Namer` interface that we had, in favor of something more go-idiomatic. Use the `linking.Name()` function instead.

server/node_kind_provider.go
```diff
@@ -0,0 +1,34 @@
...
+package server
+
+import (
+	core "typefox.dev/fastbelt"
+	"typefox.dev/lsp"
+)
+
+// NodeKindProvider consolidates the logic for gathering LSP kind information
+// based on AST nodes.
+type NodeKindProvider interface {
```
msujew: Suggestion: I would err on the side of smaller Interfaces - I.e. no completion item kind (I already have something else in mind for that). In fact, I would completely remove the `NodeKindProvider` instead and use something like the `Denominator` interface instead. Meaning that the AST node itself implements a `SymbolKind` function that returns an `lsp.SymbolKind`.

msujew (commented):
Thanks!

ssmifi:
Thanks, @msujew, this makes this actually less complex. :)
Did you think about sth like this for the `SymbolKind`?

server/symbol_kind.go
```diff
@@ -0,0 +1,27 @@
...
+	core "typefox.dev/fastbelt"
+	"typefox.dev/lsp"
+)
+
+// SymbolKindProvider can be implemented by AST node Impl structs to provide
+// custom LSP symbol kinds for the document symbol outline.
+type SymbolKind interface {
+	// SymbolKind returns the LSP symbol kind for this node.
+	SymbolKind() lsp.SymbolKind
+}
```
msujew: Minor: `SymbolKindProvider` is not a valid name anymore. Also, can we change this to `SymbolKindNode` so that the function below simply can be `SymbolKind`? That way, adopters can call `server.SymbolKind` similar to `linking.Name`.

server/document_symbol_provider.go
```diff
@@ -19,35 +20,30 @@ type DocumentSymbolProvider interface {
...
 
 // DefaultDocumentSymbolFilter provides standard symbol filtering behavior.
-type DefaultDocumentSymbolFilter struct {
-	nameProvider NameProvider
-}
+type DefaultDocumentSymbolFilter struct{}
 
 func (f *DefaultDocumentSymbolFilter) ShouldInclude(node core.AstNode) bool {
-	return f.nameProvider.GetName(node) != ""
+	return linking.Name(node) != nil
```
msujew: Suggestion: Since the default implementation of the document symbol provider only generates symbols if `linking.Name` does not return `nil`, this is redundant. Simply return `true` here?

msujew (commented):
Much better, thanks 👍

msujew (approved):
Looks good to me, thanks 👍

PR #76, opened 2026-05-21
Match lazy regexp correctly

msujew:
I've noticed during manual testing that `/* * */` broke the lexer for the multi line comment token.

This change makes the DFA/NFA generator take the greedyness properly into account.

### DIFF
--- a/internal/automatons/construct.go
+++ b/internal/automatons/construct.go
@@ -200,3 +200,23 @@ func (ck *ConstructionKit) Intersect(a, b *NFA) *NFA {
 	notAOrB := ck.Alternate(notA, notB)
 	return ck.Complement(notAOrB)
 }
+
+// LazyMatch returns the NFA matching the shortest string of the form
+// body{bodyMin,bodyMax} followed by tail. bodyMax == -1 means unbounded.
+//
+// This is the non-greedy equivalent of Concat(Repeat(body, bodyMin, bodyMax), tail).
+// The construction is:
+//
+//	greedy = repeat(body) tail
+//	result = greedy intersect complement(greedy ".+")
+//
+// A string is a minimal match exactly when it is in greedy and no proper
+// prefix of it is in greedy (otherwise the prefix-followed-by-more form would
+// place it in greedy ".+").
+func (ck *ConstructionKit) LazyMatch(body *NFA, bodyMin, bodyMax int, tail *NFA) *NFA {
+	greedy := ck.Concat(ck.Repeat(body, bodyMin, bodyMax), tail)
+	anyChar := ck.Consume(NewRuneSetFull())
+	anyPlus := ck.Repeat(anyChar, 1, -1)
+	extended := ck.Concat(greedy, anyPlus)
+	return ck.Intersect(greedy, ck.Complement(extended))
+}
--- a/internal/regexp/regexp.go
+++ b/internal/regexp/regexp.go
@@ -108,11 +108,7 @@ func newNFAFromSyntax(op *syntax.Regexp) *automatons.NFA {
 		runeSet := automatons.NewRuneSetFull()
 		return kit.Consume(runeSet)
 	case syntax.OpConcat:
-		chain := make([]*automatons.NFA, len(op.Sub))
-		for i, sub := range op.Sub {
-			chain[i] = newNFAFromSyntax(sub)
-		}
-		return kit.Concat(chain...)
+		return buildConcat(kit, op.Sub)
 	case syntax.OpAlternate:
 		alternatives := make([]*automatons.NFA, len(op.Sub))
 		for i, sub := range op.Sub {
@@ -120,15 +116,27 @@ func newNFAFromSyntax(op *syntax.Regexp) *automatons.NFA {
 		}
 		return kit.Alternate(alternatives...)
 	case syntax.OpStar:
+		if isNonGreedy(op) {
+			panic(unsupportedLazyMsg(op))
+		}
 		nfa := newNFAFromSyntax(op.Sub[0])
 		return kit.Repeat(nfa, 0, -1)
 	case syntax.OpPlus:
+		if isNonGreedy(op) {
+			panic(unsupportedLazyMsg(op))
+		}
 		nfa := newNFAFromSyntax(op.Sub[0])
 		return kit.Repeat(nfa, 1, -1)
 	case syntax.OpQuest:
+		if isNonGreedy(op) {
+			panic(unsupportedLazyMsg(op))
+		}
 		nfa := newNFAFromSyntax(op.Sub[0])
 		return kit.Repeat(nfa, 0, 1)
 	case syntax.OpRepeat:
+		if isNonGreedy(op) {
+			panic(unsupportedLazyMsg(op))
+		}
 		nfa := newNFAFromSyntax(op.Sub[0])
 		return kit.Repeat(nfa, int(op.Min), int(op.Max))
 	case syntax.OpCapture:
@@ -156,4 +164,73 @@ func newNFAFromSyntax(op *syntax.Regexp) *automatons.NFA {
 	default:
 		panic(fmt.Sprintf("unsupported syntax operation: %v", op.Op))
 	}
+}
+
+// isNonGreedy reports whether op carries the syntax.NonGreedy flag.
+func isNonGreedy(op *syntax.Regexp) bool {
+	return op.Flags&syntax.NonGreedy != 0
+}
+
+// isLazyQuantifier reports whether op is a quantifier with the non-greedy flag set.
+func isLazyQuantifier(op *syntax.Regexp) bool {
+	switch op.Op {
+	case syntax.OpStar, syntax.OpPlus, syntax.OpQuest, syntax.OpRepeat:
+		return isNonGreedy(op)
+	}
+	return false
+}
+
+func unsupportedLazyMsg(op *syntax.Regexp) string {
+	return fmt.Sprintf("non-greedy quantifier %v is only supported when followed by a fixed tail in a concatenation", op)
+}
+
+// quantifierBounds extracts the (min, max) bounds of a quantifier op.
+// max == -1 means unbounded.
+func quantifierBounds(op *syntax.Regexp) (int, int) {
+	switch op.Op {
+	case syntax.OpStar:
+		return 0, -1
+	case syntax.OpPlus:
+		return 1, -1
+	case syntax.OpQuest:
+		return 0, 1
+	case syntax.OpRepeat:
+		return int(op.Min), int(op.Max)
+	}
+	panic(fmt.Sprintf("not a quantifier: %v", op.Op))
+}
+
+// buildConcat builds the NFA for an OpConcat's children. When a child is a
+// non-greedy (lazy) quantifier, it delegates to ConstructionKit.LazyMatch to
+// produce the shortest-match construction. Recurses on the tail so multiple
+// lazy quantifiers compose.
+func buildConcat(kit *automatons.ConstructionKit, subs []*syntax.Regexp) *automatons.NFA {
+	for i, sub := range subs {
+		if !isLazyQuantifier(sub) {
+			// Skip non-lazy expressions in the regexp
+			// If there are non, simply build the concatenation as usual
+			continue
+		}
+		// Recurse into buildConcat to handle multiple lazy quantifiers
+		prefixNFA := buildConcat(kit, subs[:i])
+		tailNFA := buildConcat(kit, subs[i+1:])
+		bodyNFA := newNFAFromSyntax(sub.Sub[0])
+		minRep, maxRep := quantifierBounds(sub)
+		return kit.Concat(prefixNFA, kit.LazyMatch(bodyNFA, minRep, maxRep, tailNFA))
+	}
+
+	// Simplifications for common cases
+	if len(subs) == 0 {
+		return kit.Empty()
+	}
+	if len(subs) == 1 {
+		return newNFAFromSyntax(subs[0])
+	}
+	// Simply concatenate all subexpressions
+	// This is the default case if there are no lazy quantifiers
+	chain := make([]*automatons.NFA, len(subs))
+	for i, sub := range subs {
+		chain[i] = newNFAFromSyntax(sub)
+	}
+	return kit.Concat(chain...)
 }
--- a/examples/statemachine/lexer_gen.go
+++ b/examples/statemachine/lexer_gen.go
@@ -353,7 +353,7 @@ var Token_ML_COMMENT = core.NewTokenType(
 	Token_ML_COMMENT_Idx,
 	"ML_COMMENT",
 	"ML_COMMENT",
-	core.SkippedGroup,
+	core.CommentGroup,
 	core.TokenKindToken,
 	0,
 	false,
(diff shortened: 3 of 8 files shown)


internal/automatons/construct.go
```diff
@@ -200,3 +200,23 @@ func (ck *ConstructionKit) Intersect(a, b *NFA) *NFA {
...
+//
+// A string is a minimal match exactly when it is in greedy and no proper
+// prefix of it is in greedy (otherwise the prefix-followed-by-more form would
+// place it in greedy ".+").
+func (ck *ConstructionKit) LazyMatch(body *NFA, bodyMin, bodyMax int, tail *NFA) *NFA {
+	greedy := ck.Concat(ck.Repeat(body, bodyMin, bodyMax), tail)
+	anyChar := ck.Consume(NewRuneSetFull())
+	anyPlus := ck.Repeat(anyChar, 1, -1)
+	extended := ck.Concat(greedy, anyPlus)
+	return ck.Intersect(greedy, ck.Complement(extended))
```
Lotes, line 221: Math power :D

internal/regexp/regexp.go
```diff
@@ -157,3 +165,66 @@ func newNFAFromSyntax(op *syntax.Regexp) *automatons.NFA {
...
+	panic(fmt.Sprintf("not a quantifier: %v", op.Op))
+}
+
+// buildConcat builds the NFA for an OpConcat's children. When a child is a
+// non-greedy (lazy) quantifier, it delegates to ConstructionKit.LazyMatch to
+// produce the shortest-match construction. Recurses on the tail so multiple
+// lazy quantifiers compose.
+func buildConcat(kit *automatons.ConstructionKit, subs []*syntax.Regexp) *automatons.NFA {
+	for i, sub := range subs {
+		if !isLazyQuantifier(sub) {
```
Lotes, line 209: I do not understand this.
You want to concat multiple RegExps but exit after the first lazy quantifier?
...
Ah wait there is a recursion... reads a bit hard. Is it possible to simplify the flow? What is happening here?

internal/regexp/regexp.go
```diff
@@ -157,3 +165,66 @@ func newNFAFromSyntax(op *syntax.Regexp) *automatons.NFA {
...
+		return int(op.Min), int(op.Max)
+	}
+	panic(fmt.Sprintf("not a quantifier: %v", op.Op))
+}
+
+// buildConcat builds the NFA for an OpConcat's children. When a child is a
+// non-greedy (lazy) quantifier, it delegates to ConstructionKit.LazyMatch to
+// produce the shortest-match construction. Recurses on the tail so multiple
+// lazy quantifiers compose.
+func buildConcat(kit *automatons.ConstructionKit, subs []*syntax.Regexp) *automatons.NFA {
```
Lotes, line 207: 1.
"buildConcat" does something because of lazy/greedy context.
"kit.Concat" joins two NFAs.
Can you reflect what is the bug difference in its name. Currently both sound equal.

2. Might it be an option to include this build method in the construction kit? I guess not because the kit refers to NFAs and buildConcat refers to Regexps....

internal/automatons/construct.go
Lotes: greedy = B{min,max} Tail
anyPlus = .+
extended = B{min,max}Tail.+ = B{min}Tail.+ | ... | B{max}Tail.+
result = greedy \ extended // greedy without extended

It is a bit hard to understand for me why this is lazy match.
I was wondering how to implement laziness using NFAs when I was writing the Regexp part. Is this really the right way?

Lotes (commented):
Can you really implement laziness using NFAs? It is a bit long ago. I am not sure if I had it in my lectures.

internal/automatons/construct.go
msujew: The code is rather obscure. The Idea is to turn:

```
L = X{a,b}?·T
```

Into:

```
L = X{a,b}·T ∩ ¬(X{a,b}·T·Σ⁺)
```

I.e. turn the non-greedy part into a language that can only match the original language **and** the complement of the same language followed by any characters. I think the comment on `LazyMatch` explains this well enough:

 The focus is on the "shortest string" part. The alternative would be that the regexp sees `*/`, matches the `*` inside of the greedy loop and then continues to `/`, which theoretically brings it to an accepting state. But, because there is still input to be greedily matched afterwards, it does this until it arrives at the end of the input, where it then returns "no match".

At least that's my understanding of it. To be honest, I didn't come up with this, but it seemed reasonable, so I just went with it ¯\\\_(ツ)\_/¯

internal/regexp/regexp.go
msujew: The flow is:
1. Either we find a lazy expression, in which case we split the NFA into prefix, tail and body, on which we then call the `LazyMatch` constructor method. The recursion is there to help deal with subexpressions that are also lazy
2. We don't find one, and so we simply fall back into the normal `Concat`.

I don't think we can reasonably simplify this - but I'll add additional comments for this.

internal/regexp/regexp.go
msujew: I wanted to have this in the construction kit as well, but we need the `syntax.Regexp` object to identify the greedyness. That's why it has to live here.

Note that by default, `buildConcat` will perform an NFA `Concat` operation. Just in the lazy case, it will also call `LazyMatch` in addition to that.

msujew:
> Can you really implement laziness using NFAs? It is a bit long ago. I am not sure if I had it in my lectures.

I'm 100% sure that I didn't have this in my lectures as well. It turns out yes, but not completely. That's why the constructor panics if lazy parts of the regexp are nested inside of non-lazy groups. We probably need some way to validate this in the language itself, before converting it into the NFAs (where it will panic the generator).

internal/automatons/construct.go
Lotes: I made a [node editor](https://simulations.language-benders.io/?share=tZdLb9s4EMe_i467FsHH8DG-toftYXsqFgs0QSHEiiPAkQNL7jYb5Lvv0JIcySYdql37YBgyNfz_OE--ZPV2VTbZ8utLVq2yZfZU7JoyF9YiorEaBPBcZousfX4qh78_0yv06GnbVG21rbPlS_YjW2qHzAl0SG8h_TCL7DlbSjRMO8VRoBJKgH1dZKuiLegl-vVYFs1-V668iX-qVfuQLYXhi-yhrNYPrbdJi5pyU961ftF9sWlKen9XrNdVve4fvC567UXT5vV90at3xiIYlfM39bTgy_bzfXEu3qJmXqHV9AGr5VE8gEYO1tGXuKL2pt3R0146CsmdmEjv_v9UP-3bc_XSSmYFImjQRjs6706-5sxYIx1qlBz1SH_2vdjsvd2_f_vy9eZmdXPz8fb3LEwl-YiKlP0UluNSSiFRczkDSzHpwEoNEqVUqA9YFhg6a7VAabkltDDWFWiG5OhgKE6EzEVickhmOFCOCKskJdaBRHDNtBDWGMXpcJyCK0bYW3Z08q3w8mVKdtALzDgH3CkrOb3XqwdmgSoEGlBcS3BXVO-VV3Vb0gnftQMDBTidXa7eGGjZp2HVOYeQFpkS3ArQVtDpuw4EDKOwkESBSCXPjiOq8rH5YbuvSbJMwJoZUp7rbvv4tCkfy3oE5rjhOUzAPhyXBcg4KEb11X_IQF98FedMak21RChHycevWX09yfeqOSJYDlzneoLwV9Xsi031b7kLIGiNjIPTmpoH5brrGIQWzFDYGc0VdRcwsxCcmCC0u306gVLUQWRuZhAoKk2GU7MQIKj5HAByoxlyIbx-B6DUvDZyQpDghFV5jCNltDIutxOEjyUlyGNVE0QwkCz1POm4NoZzC50TDHOAkto6-ZQ7o9Qshl8KJOBKC5O7GW4AZIjUEJVR2vRuUMgZKIPSEBe12HkE871AB3wkoHjgLscJwZ_kgMewCyQCU74lWEAquUMcCZ8eijiojRjBcV4mzPfBKI5QUSSIcbNIiCOSSz2bphKaBa20BwhQ6Js3UAtE439cmWHkBQSq_PykWVzwAp0_czQ1ARE4ajRdy9bUK9DRQGKpMIGdWY7mA4wSgQY46l8nTeFyIliKeU2V2HOQA7tk1tYSmLXKOZqqrLhGJtwusnK17i8WzXa_uzsOSIHLRbfgj6JebfpBJN8eBsK22K3LNrs42h-WTN6tanp-OMEfz7mX8e1bcOd-nzxivDfl3XEkiI7pJwj9uhOK6OVqytC_HMAIb_-2WR7c4c1emOV0Np_FcjYLz2aZbB9gmewQYYmpmRNZZ2PxrMia7HweWRPjociKR_gJg190wvD-EDlF6YbpM5SIhH7D_PIuvc0I0-nZpjFdHPhDSIfzikFNNIyhotscjU6o3j3ta8GJM7j3XZLKeEiWaVU_g76kOI05cj9Iis53KQbUwCbXCM7ImJ3EkhCUAfMhCv8sPGqmUUTuO0kUkb0jvhjMxyjC55lGET6BdIrIYQ8UEcgQxf9QGCJ3n2SYtIIQ2OVylvxcW4pcIpJgEvpRwHxCfI2UpFGMAmB8jUiiuCAzEF-D-RhFWEkaReQukUwRkRmIqMH8QHH7-h8) for playing around with NFA operators. During the development there occured some bugs. It can be likely that the same bugs are in the current automaton implementation of Fastbelt. I added the current scenario of this discussion to the given link.
Note that .* is given as [\d\D]. Still need to make some things „round“.
However, looking at the green boxes, the NFA expression with the intersection has not the desired effect. It is the same as the first operand.

AFAIK lazyness is implemented by the runner of the DFA/NFA. I looked at [here](https://swtch.com/~rsc/regexp/regexp2.html), where a Regexp is implemented using a virtual machine. Lazyness was implemented there by prioritizing the threads of the SPLIT opcode. Either the first address it jumped to first (greedy) or the second address (lazy). Lets see if I am wrong or we need to fix stuff.

Lotes: My simulation was not detailed enough (here is a better [one](https://simulations.language-benders.io/?share=xZjNbuPGEoXfhctE7FRV10-XtpPFzSLZ3Its4kFAmLRHgE0ZEj3JZOB3D2jJEik3ZWnuaGZlgLS662Od06zDz8X6-kNzX_3erNaLZVvMeVa0y7pZF_M_PheLupgXD9Vq3ZRo5u5qwshQUjErHpbrRff8m8_F38VckgRgE8eIFhHSrPhUzAUDgyRxiopmKE-zovv00Lys-9uybopZUVddVcw_P82K-6ZaP66aul_1r0XdfSjmqDArPjSL2w9dMVd4mhXr5q657vp_uqnu1s2sqFfV7e2ivd1eeJpta6_WXdneVNvqk5qzxhJeV28OgQmSeiRyVX6pnjSpEXh0iBz31Vfr7n_L326qCxa_7laL9nZbuyNBwmztpBacTaMTMZi6bYoP7A6iydWl79u--M3Kv7QPj92-_uJjdffY3_3j6qq-uvr5_Q9FHolggISEX8SUgIiQXIDyTCkYgptHjMn0mYjIggD1z0IAXaKcinR19cNPF6B5scYGRtiQSsxYw2IAcTc1SCQbFqXgbsnQjR0t6vdyxqZ4w774jK8TWGAFQ1ZgIUnb8iUwojklYEyg-k2t0Vd-vWyvq-4ZgE1VSFhzUkLwFBIYKhGoMz0DIEpg0ChJjEBEB1pqb6p3z4sPlLTolfVu-dh2xZxO4EE6C2ivJDa16AYTh2yPIpJSUoItCrMHJqVELgrM9G2VNLB0XzuLmJcxY2mzYASYKIEJ-kZIjClY9JgUADSxxHOPqR8vYOyhO3ooA4pWcu69YSEyJzRLCW1rbpYYVARUDFkAgL-zO8wJEpeScUfUGDiaq0JEdN0QxMgBhGL_3gaJiN_VHfVNVdbNjiYxakylZmgkaVAEZPNEiLh5j0eRIOQeWfshZnhY1TfVz03XrO4X7eKfS9qkfm7K_cMOQtAYSstAmEsgckwoiSFtO6IQEhObWTKmaCOGd8v7h7vmvmm7CyMM26AcE5Qp2wYKFCMliODq22GQgEIyRSWGfpqK36kNi7ZrVuvmek-S3NlLzxxaiBwQXYgFKJnEDUmUoMLKYqaiPm7GLy_rfzuHfFysX2AcopOUmJuqjC0wRzGMLFtzlIhBWJyAzYUEh9L6uFg_VneLf5rV8ATeXfxv36w9gWEMgMncxShF5z2PKARHQkaSmJTxKf8QDOOgtWjn9vZ-0e4eBBOqlpiZyIgFg4EDW7Rex7uhRtwjqKbefzLq6q-LdnF_eXHuAQyTAlGJmUGAEGJwQOR-qEnxpZUpBWFLQppMyOSbI-yFaOhOqCVmZwFOAdz6v9FtO1SWan0Kc3HgBJaEv1iIIAHForIbodpAh8JBIrtgVGYyn9IhyFCHJ00P72dFU98Os_Pfn8r-yp9_ZkN0P2UsH7tyIqX2lxdtMSvWy8fVdXMkinfV6rbpjubdzRr_qdr6bjuF9Fvvfju6sdl0TLtr8A4pH0-3V3usbLXb-2OuyaS7A5siP8Da7_6abLTzyXCjnJqBG0W_43CHkfcV3GGMvBRcdr_XYhwFw2kxHladEeNhxPz6YpwQfn-ph5rOi-3zPDCwyFEX7diOBtADvG0Rr_GGe58OOHqYJwNiHvCwMxcBxDMdN4iRWccNIvJbjhsn0ozjxnH7ixx3zGCD5XMGG2TLYwYbF5k12Dilnmmw43IbLD0tt5eE-ZbcxnUekdsgsp4tt9c8x_1xEtLoiDhqj4tQwUlUw5p7qonkvFnzONCg1B3QdBA_jybDMrF0PQDJxef6Fcl0jSOMiSx-wFFPcdRTHG8rbSJGn9KTnMimU_lX78l26WFPJrN0nXHNdKkjmmP5_Lz-ZF1zTEqnYuEh1oScLsWVO-Om4u8LV-4bQdV-ythnIkTvYCa-NpzKsN00T3D0yb-0ZwrzNcqxZz5qzgTz_30enOChXOR_-1ybMM_E94OvwjFV6Fhd--D_lrpGFWbUNfqEcK663j_9Cw)). You are simulating a lazy expression with a greedy execution mode by reshaping the DFA.

It is a bit weird that it is only covering this case in combination with concat. There are these four of 5 switch-cases that are panicing now (5 of 5 would be the previous state, so it is actually an improvement).

I am wondering if we can do the same for the other 4 cases? The attractive outcome is that we are using DFAs then, so only one execution thread. Lazy/Greedy paths can only be decided in an NFA or a VM, which would run longer because of "multithreading" all paths (until the first one matches).

In terms of complexity, I do not like the solution so much. Can we remove it when there is a lazy execution possible? For me it was hard to understand: 1 of 5 cases plus the crazy recursion when building the concat.

internal/regexp/regexp.go
Lotes: Rereading this, I think the organization of code is fine

Lotes (approved):
Good solution. I hope we can fix the other lazy cases somewhen as well.

PR #77, opened 2026-05-21
Publish VS Code extension

spoenemann:
Added README.md, LICENSE, and a publishing script for cross-compiling for multiple platforms. This is meant as a starting point – we can extend the available platforms later and can also consider migrating to another scripting language or a GitHub Action.

### DIFF
--- a/internal/vscode-extensions/fastbelt/publish.sh
+++ b/internal/vscode-extensions/fastbelt/publish.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Require tokens for both marketplaces before doing any work.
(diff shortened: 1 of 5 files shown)


internal/vscode-extensions/fastbelt/publish.sh
msujew, line 1: Suggestion: Since `node` is required anyway to run this script, can we convert this into a JS script? That way, it can also run on Windows ;)

msujew (commented):
Makes sense. I have one comment though.

msujew (approved):
Nevermind, you mentioned this in the PR description. We can merge it as is and later update it to use a different scripting language - fine with me.

PR #78, opened 2026-05-22
Documentation for root package and grammar

spoenemann:
Added a root package doc to be shown here: https://pkg.go.dev/typefox.dev/fastbelt

This explains the basics of the framework and delegates to subpackages for further details (those need to be documented as well).

Also added a `grammar` package with the sole purpose of documenting the grammar language. We can't do that in `internal/grammar` because that's not public (doesn't need to be – framework users don't interact with the grammar directly).

@msujew please check whether the grammar reference is complete and correct wrt. the current state of the implementation.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,0 +1,132 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+// Package fastbelt is a language engineering toolkit for Go.
+//
+// # Overview
+//
+// Fastbelt covers lexing, parsing, AST creation, cross-reference linking,
+// workspace handling, and Language Server Protocol (LSP) support.
+//
+// Fastbelt is primarily inspired by [Xtext] (Java)
+// and [Langium] (TypeScript).
+//
+// Like these frameworks, Fastbelt uses a grammar definition file as the
+// entry point, based on a Grammar language that is itself implemented with
+// Fastbelt. See [typefox.dev/fastbelt/grammar] for details.
+//
+// The [typefox.dev/fastbelt/cmd/fastbelt] command generates Go code from a
+// grammar definition, which you can integrate and customize with additional
+// code.
+//
+// # Scaffolding
+//
+// The [typefox.dev/fastbelt/cmd/fastbelt] command can scaffold a new language
+// project with:
+//
+//	fastbelt scaffold -module example.com/you/mylang -language "MyLanguage"
+//
+// The scaffold creates these root files in the generated package directory:
+//
+//   - gen.go with go:generate directives for grammar-based code generation.
+//   - services.go with customization points for generated services.
+//   - <language-id>.fb as the initial grammar definition file.
+//   - cmd/<language-id>-lsp/main.go as the LSP server entrypoint.
+//
+// By default, scaffolding also creates VS Code extension files in a
+// `vscode-extension` subfolder.
+//
+// After writing templates, scaffolding runs go generate and go mod tidy so the
+// generated parser, lexer, linker, and supporting files are ready to use.
+//
+// # Defining the Grammar
+//
+// Defining a language in Fastbelt is an iterative design process. You
+// co-design the concrete syntax that users write and the abstract syntax tree
+// (AST) that your tooling consumes.
+//
+// In practice, you write interface declarations for AST node types and parser
+// rules that describe how instances of those node types are created from input
+// text. Each assignment in a parser rule maps matched input to a field on the
+// node that the rule builds.
+//
+// Parser rules always consume contiguous regions of text. They match keywords
+// and tokens directly and delegate to other rules for nested regions. This
+// keeps grammar structure aligned with language structure and makes it clear
+// how parsed text becomes AST subtrees.
+//
+// Cross-references provide built-in linking from symbol usage to symbol
+// definition. Instead of creating a new node, a cross-reference records an
+// identifier that the linker resolves to an existing AST node in scope.
+//
+// Because Go interfaces are generated directly from grammar interfaces, these
+// interfaces should reflect how you plan to implement semantic checks,
+// interpreters, or code generators. At the same time, there's a 1:1
+// correspondence between interfaces and parser rules so the parse model and
+// runtime model stay easy to reason about.
+//
+// The grammar language itself is documented in
+// [typefox.dev/fastbelt/grammar], while the code generation workflow is
+// documented in [typefox.dev/fastbelt/cmd/fastbelt]. A common workflow is to
+// evolve the grammar in small steps and run code generation after each change.
+//
+// # Customization
+//
+// There are two ways to add custom behavior to your language:
+//
+//   - Attach behavior directly to specific generated AST node `Impl` types.
+//   - Implement custom services and wire them through the service container.
+//
+// Validations are implemented with the first approach. To add validation checks
+// for a node type, implement [Validator] on that node's generated `Impl`
+// struct:
+//
+//	type PersonImpl struct {
+//	    PersonData
+//	}
+//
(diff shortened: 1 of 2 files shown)


grammar/doc.go
```diff
@@ -0,0 +1,329 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+// Package grammar documents the Fastbelt grammar language.
+//
+// A grammar definition file (extension .fb) describes the concrete syntax
+// and type structure of a language. Fastbelt reads .fb files and generates
+// Go lexers, parsers, and AST types from them. The grammar language is
+// itself implemented as a Fastbelt grammar.
```
msujew: Suggestion: Since the specific code generated by the cli is changing sometimes, I would keep this a bit more generic like "generates Go code".

grammar/doc.go
```diff
@@ -0,0 +1,329 @@
...
+// A token rule matches a character sequence with a regular expression and
+// assigns it a name. Token rule names are conventionally written in upper
+// case:
+//
+//	token ID:  /[_a-zA-Z][\w_]*/;
+//	token INT: /[0-9]+/;
+//
+// The order in which token rules are declared is significant: the lexer
+// returns the first match, so more specific patterns should appear before
+// more general ones.
```
msujew: Suggestion: That's not *quite* correct. It returns the first, longest match. I.e. if two tokens match the same text (exact same length), then the first defined token wins.

grammar/doc.go
```diff
@@ -0,0 +1,329 @@
...
+// # Comment Tokens
+//
+// Documentation comments attached to the grammar element that follows them
+// are declared with the comment modifier:
+//
+//	comment token SL_COMMENT: /\/\/[^\r\n]*/;
+//
+// Comment tokens are moved in front of the element they precede during
+// parsing, making them available for tooling such as hover documentation in
+// a language server.
```
msujew: What is this supposed to mean? This only makes sense in the presence of the CST like in Langium. In Fastbelt, we store them inside of a `Comments []Token` slice attached to the document where they can be accessed.

grammar/doc.go
```diff
@@ -0,0 +1,329 @@
...
+// parentheses can carry a cardinality:
+//
+//	Model returns Model:
+//	    (Persons+=Person | Greetings+=Greeting)*;
+//
+// # Keywords
+//
+// A keyword is a literal string in double quotes. Keywords guide the parser
+// and provide visible structure to the language. They must not be empty and
+// must not contain whitespace:
```
msujew:
```suggestion
// and provide visible structure to the language. They must not be empty.
```

grammar/doc.go
```diff
@@ -0,0 +1,329 @@
...
+//   - composite -> getter returns `string`; scalar composite fields also get a
+//     `FieldNode() core.CompositeNode` accessor.
+//   - TypeName -> getter returns `TypeName`.
+//   - *TypeName -> getter returns `*core.Reference[TypeName]`.
+//   - []FieldType -> getter returns a slice of the mapped element type.
+//
+// Slice fields use append-style setters (`Set<Field>Item`) in generated code.
+// Scalar fields use `Set<Field>(value ...)`.
+//
+// # Token Rules
```
msujew, line 103: Suggestion: Maybe also note that keywords are always attempted to be matched before any explicitly declared token.

grammar/doc.go
```diff
@@ -0,0 +1,329 @@
...
+// Boolean assignment (?=) sets a bool field to true when the right side is
+// consumed; the field remains false otherwise:
+//
+//	Employee returns Employee:
+//	    "employee" Name=ID (Remote?="remote")?;
+//
+// Assignments with cardinality + or * form a contiguous group: the sequence
+// of matched values must not be interrupted by elements belonging to a
+// different assignment before the group is complete. Hidden tokens such as
+// whitespace and comments may appear anywhere without interrupting a group.
```
msujew: The part about hidden tokens doesn't really fit to `Assignments` - maybe to `Groups` above, or just directly stick this to the hidden tokens part?

grammar/doc.go
```diff
@@ -0,0 +1,329 @@
...
+//
+// # Cross-References
+//
+// A cross-reference reads an identifying token from the input and resolves it
+// to an existing object rather than creating a new one. The syntax is:
```
msujew: Theoretically, a Reference token can be used as a `SymbolDescription` declaration or a `Denominator` return value - meaning that it could "create a new one".

grammar/doc.go
```diff
@@ -0,0 +1,329 @@
...
+//
+//	property=[Type:TOKEN]
+//
+// Type is the name of an interface and TOKEN is the name of a token rule that
+// identifies objects of that type. If TOKEN is omitted, Fastbelt uses the
+// token matched by the Name field assignment of the referenced type:
```
msujew: As of #55, composite rules can also be used as cross reference items.

grammar/doc.go
```diff
@@ -0,0 +1,329 @@
...
+//	    Definition | DeclaredParameter;
+//
+// The parser rule AbstractDefinition does not create an object of its own.
+// Instead it calls either Definition or DeclaredParameter, and whichever
+// rule matches creates the object. This pattern is the standard way to write
+// rules that match one of several concrete types.
+//
+// In contrast, an assigned rule call such as parameter=DeclaredParameter
+// creates an object in the current rule and assigns the result of the called
+// rule to the named property.
```
msujew, line 284: Suggestion: Add that after consuming an unassigned subrule, additional properties can be assigned to the AST node - but not before consuming it.

grammar/doc.go
```diff
@@ -0,0 +1,329 @@
...
+// When the "+" keyword is found, a new Addition object is created, the
+// object parsed so far is stored in its Left property, and that new Addition
+// becomes the current object. The operator += is also valid for tree-rewriting
+// actions on slice properties.
+//
+// # Composite Rules
+//
+// A composite rule matches a structured token value such as a qualified name
+// or dotted path. Unlike parser rules, composite rules do not create AST
+// objects; they yield an opaque composite value that is stored in a field of
```
msujew: Why _opaque_? It can be inspected by looking at its `Tokens()` method to see each individual token.

grammar/doc.go
```diff
@@ -0,0 +1,329 @@
...
+//
+// Declare the receiving field as composite in the interface, then use a
+// normal = or += assignment on the right side of a parser rule:
+//
+//	interface TypeRef {
+//	    Name composite
+//	}
+//
+//	TypeRef returns TypeRef:
+//	    Name=QualifiedName;
```
msujew: Better example:

```suggestion
//	interface Type {
//	    Name composite
//	}
//  interface TypeRef {
//      Item *Type
//  }
//
//	Type: Name=QualifiedName;
//  TypeRef: Item=[Type:QualifiedName];
```

doc.go
```diff
@@ -0,0 +1,132 @@
...
+// keeps grammar structure aligned with language structure and makes it clear
+// how parsed text becomes AST subtrees.
+//
+// Cross-references provide built-in linking from symbol usage to symbol
+// definition. Instead of creating a new node, a cross-reference records an
+// identifier that the linker resolves to an existing AST node in scope.
+//
+// Because Go interfaces are generated directly from grammar interfaces, these
+// interfaces should reflect how you plan to implement semantic checks,
+// interpreters, or code generators. At the same time, there's a 1:1
```
msujew: Instead of _1:1_ I would say _close_, since a rule can change the type of the return value (actions), which makes it no longer 1:1.

```suggestion
// interpreters, or code generators. At the same time, there's a close
```

msujew (commented):
Thanks, I have a few comments on the grammar language documentation.

msujew (approved):
Thanks!

PR #79, opened 2026-05-26
Support completion LSP feature

msujew:
Adds support for Langium-style code completion - it's a bit more structurally sound compared to Langium:

1. Since we cannot purely rely on runtime reflection for certain features (such as synthetic AST nodes, if a reference appears as its first feature), this change adds a `completion_gen` file that helps us with some of the reflection handling. In Langium, we were able to do this inline.
2. Implements ATN simulation for the completion use case. This is a sounder approach than what we did in Langium, where we had to interpret the current grammar at runtime.
3. Generates a new completion parser as well. We need to track a lot of ATN intermediate information for the completion. While we theoretically could track this during the normal parse as well, this would slow down the normal parser, which is undesirable.
4. Based on top of the error recovery work, otherwise most of these features wouldn't work at all (since completion operates on incomplete input).
5. Adds 2 ways to extend the functionality of the completion provider:
  1. A `CompletionContributor` works similar to the common override of the `CompletionProvider` in Langium - specify exactly how a completion item is generated from a specific source token type and ATN state. This also works for cross references and snippets/templates.
  2. A `CompletionFilter` that allows more fine-grained control over which `SymbolDescription` elements are returned as part of the cross-reference completion logic. Acts similar to the existing `LanguageScopeProvider`.

This is quite an extensive change, but it's also extensively tested. Also fixes a bunch of bugs in the error recovery, mostly related to `nil` handling at EOF.


cmd/fastbelt/main.go
```diff
@@ -160,11 +160,19 @@ func runLegacyGenerate(args []string) error {
 		generator.GenerateTypes(grammar, packageName)); err != nil {
 		return err
 	}
+	tokenTypes := generator.GenerateTokenTypes(grammar)
 	if err := writeFile("parser", filepath.Join(outputPath, "parser_gen.go"),
-		generator.GenerateParser(grammar, packageName)); err != nil {
+		generator.GenerateParser(grammar, packageName, tokenTypes)); err != nil {
+		return err
+	}
+	if err := writeFile("completion-parser", filepath.Join(outputPath, "completion_parser_gen.go"),
```
spoenemann, line 168: Completion is an LSP concept, so we should generate this to a different package than the base language impl. The default could be a "server" subpackage.

spoenemann (commented):
Works great!

cmd/fastbelt/main.go
msujew: I'm not entirely on board with this. It leads to a lot of complexity in the generator and the adopter code.

Instead, what do you think about splitting the `SetupGeneratedServices` into a normal `SetupGeneratedServices` and a `SetupGeneratedServerServices`? That way, only adopters actually calling `SetupGeneratedServerServices` have to include the LSP specific code into their binary.

spoenemann: Yes that works.

> It leads to a lot of complexity in the generator and the adopter code.

Could you explain why?

msujew: Mainly two reasons:

1. Right now, our convention is to use `./server` as the binary, meaning that its package name is `main`. However, the generated package name should probably be `server` instead - or should we move the main language server binary somewhere else (by convention)?
2. I personally feel like it's much clearer when a CLI just outputs files into a single directory and doesn't start creating subdirectories somewhere else. That way, adopters don't have to "compete" with directories created by our CLI.

spoenemann (approved):
This is a great starting point. We can refine it over time where needed.

PR #81, opened 2026-05-27
Import only grammar symbols from same package

spoenemann:
I propose that grammars can only see definitions from the same package. This keeps the Fastbelt grammar aligned with Go. At a later stage, we can evaluate adding explicit `import` statements, potentially with the same syntax as Go.

Also two renamings to make service names simpler and shorter:
- ImportedSymbolsProvider → SymbolImporter
- ExportedSymbolsProvider → SymbolExporter

I kept LocalSymbolsProvider as the analogy doesn't work there.

### DIFF
--- a/internal/grammar/imported_symbols.go
+++ b/internal/grammar/imported_symbols.go
@@ -1,0 +1,41 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+package grammar
+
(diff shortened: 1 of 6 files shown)


msujew (commented):
How does the CLI deal with this? Should it load all `.fb` files in the same directory as the target file?

spoenemann:
Yes; I created #83 for this.

msujew (approved):
Thanks, looks good to me 👍

PR #82, opened 2026-05-27
Use Cobra CLI lib, add generate command

spoenemann:
This PR adds [Cobra](https://pkg.go.dev/github.com/spf13/cobra@v1.10.2) for a proper CLI incl. usage messages etc.

It also introduces an explicit `fastbelt generate` command.


msujew (commented):
I believe the `launch.json` hasn't been updated?

spoenemann:
Right, updated that.

I also propose two breaking changes:
- Make the grammar file an argument of the `generate` command instead of an optional flag
- Make the vscode extension default to false in the scaffold command – I wouldn't push that feature too much in this case

Ok with that?

cmd/fastbelt/main.go
```diff
@@ -102,10 +101,10 @@ func newScaffoldCmd() *cobra.Command {
...
 		},
 	}
-	cmd.Flags().StringVar(&opts.modulePath, "module", "", "Module path for go mod init (optional; omit to scaffold into an existing module)")
-	cmd.Flags().StringVar(&opts.packagePath, "package", ".", "Template output directory: with --module, relative to the new module root; without --module, relative to the working directory")
-	cmd.Flags().StringVar(&opts.language, "language", "", "Human-readable language name")
-	cmd.Flags().BoolVar(&opts.noVSCodeExtension, "no-vscode", false, "Do not generate a VS Code extension")
+	cmd.Flags().StringVarP(&opts.modulePath, "module", "m", "", "Module path for go mod init (optional; omit to scaffold into an existing module)")
+	cmd.Flags().StringVarP(&opts.packagePath, "package", "p", ".", "Template output directory: with --module, relative to the new module root; without --module, relative to the working directory")
+	cmd.Flags().StringVarP(&opts.language, "language", "l", "", "Human-readable language name")
+	cmd.Flags().BoolVar(&opts.createVSCodeExtension, "vscode", false, "Generate a VS Code extension")
```
msujew, line 107: Suggestion: The default for `createVSCodeExtension` should be `true`

internal/e2e_test/scaffold_e2e_test.go
```diff
@@ -57,7 +57,7 @@ func TestScaffoldModule_endToEnd(t *testing.T) {
 	})
 
 	t.Run("scaffold module", func(t *testing.T) {
-		scaffold := exec.Command(fastbeltBin, "scaffold", "-module", modulePath, "-language", "E2E Lang")
+		scaffold := exec.Command(fastbeltBin, "scaffold", "--module", modulePath, "--language", "E2E Lang", "--vscode")
```
msujew, line 60: Suggestion: Related to the above - we discussed in #59 that we want to use opt-out instead of opt-in for the vscode extension. Do you want to change that now?

msujew (commented):
Thanks, looks quite good. You'll have to rebase, now that #79 is merged.

internal/e2e_test/scaffold_e2e_test.go
spoenemann: Oh I didn't see that discussion. Yes my proposal is to change that, as it feels cleaner to me to generate just a Go project by default. But I don't insist on it.

msujew: Fine with me 👍

msujew (approved):

PR #84, opened 2026-05-29
Added comments to public definitions in root package

spoenemann:
Added lots of missing doc comments to public definitions.

### DIFF
--- a/ast.go
+++ b/ast.go
@@ -372,8 +372,7 @@ type NamedCompositeNode interface {
 	NameNode() CompositeNode
 }
 
-// StringUnit is a common interface for both [Token] and [CompositeNode], as both can serve as
-// the "name" of a reference.
+// StringUnit is a common interface for both [Token] and [CompositeNode].
 type StringUnit interface {
 	// Owner returns the AST node that owns this string unit.
 	Owner() AstNode
--- a/ast.go
+++ b/ast.go
@@ -10,6 +10,59 @@ import (
 	"sync/atomic"
 )
 
+// AstNode is the base interface for all AST nodes.
+//
+// Every language-specific AST node type which is generated from a grammar definition embeds
+// this interface.
+type AstNode interface {
+	// Document returns the owning document of the node.
+	Document() *Document
+	// SetDocument sets the owning document of the node.
+	//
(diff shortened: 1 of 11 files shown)


ast.go
```diff
@@ -286,59 +349,76 @@ func AssignContainers(doc *Document, root AstNode) {
...
+// stored in the "Name" field of the node.
 type NamedCompositeNode interface {
 	NamedNode
+	// NameNode returns the composite node stored in the node's "Name" field.
 	NameNode() CompositeNode
 }
 
-// [StringUnit] is a common interface for both [Token] and [CompositeNode], as both can serve as the "name" of a reference.
+// StringUnit is a common interface for both [Token] and [CompositeNode], as both can serve as
+// the "name" of a reference.
```
msujew: Suggestion: While originally, this was only for references, this has actually evolved to be used in the `SymbolDescription` and other places. You can remove the part after the comma.

document.go
```diff
@@ -16,51 +16,89 @@ import (
...
-	References            []UntypedReference
+	// URI identifies the document in the workspace.
+	URI URI
+	// State tracks which build phases already ran for this document.
+	State DocumentState
+	// Root is the AST root produced by parsing.
+	// It is nil until parsing succeeds.
+	Root AstNode
+	// Tokens is the token stream produced by lexing.
+	// It is empty until lexing succeeds.
```
msujew: Suggestion: Not a stream.

reference.go
```diff
@@ -16,25 +16,11 @@ import (
...
-	Text() string
-}
-
-type ReferenceGetter[T AstNode] func(context.Context, *Reference[T]) (*SymbolDescription, *ReferenceError)
-
 // Reference represents a reference to another AST node of type T.
-// Resolving is thread safe and is done concurrently by default.
-// The resolution is triggered when [Ref], [RefNode] or [Resolve] are called for the first time.
+//
+// Resolving a reference is thread safe and is done concurrently by default.
```
msujew: Suggestion: This used to be the case initially, but since then the behavior has changed - the `Linker` is responsible for the concurrent behavior. The `Resolve` function is now single-threaded. Nevertheless, calling `Resolve` and other function is thread-safe.

msujew (approved):
Looks good 👍

PR #85, opened 2026-05-29
Add workspace symbol feature

### LINKED ISSUE #29: Workspace symbols
Related to https://github.com/TypeFox/fastbelt/issues/25. Note that this does not rely on https://github.com/TypeFox/fastbelt/issues/30.

ssmifi:
Fixes https://github.com/TypeFox/fastbelt/issues/29
and implements the workspace symbol provider, which is basically a port from Langium.

Customization is aligned to the recent providers. 
There is also a suggestion to include a max symbol count, but we can drop if desired. 

Also, I noticed that the completion service added fuzzy matcher as a service. In this approach similar to before I tried to simplify the interface and just added a fuzzy function with the same functionality as a suggestion. But I can reuse the matcher, if you think it is worth it in the long run.


server/workspace_symbol_provider.go
```diff
@@ -0,0 +1,137 @@
...
+
+		if maxCount > 0 && len(symbols) >= maxCount {
+			return symbols[:maxCount], nil
+		}
+	}
+
+	return symbols, nil
+}
+
+func (p *DefaultWorkspaceSymbolProvider) collectSymbolsFromDocument(doc *core.Document, query string, currentCount, maxCount int) []lsp.SymbolInformation {
```
msujew: Suggestion: Instead of returning the array here, and then re-adding it to the other array, can we accept the main array as an argument and the elements there directly?

server/workspace_symbol_provider.go
```diff
@@ -0,0 +1,137 @@
...
+		filter: filter,
+	}
+}
+
+func (p *DefaultWorkspaceSymbolProvider) HandleWorkspaceSymbolRequest(ctx context.Context, params *lsp.WorkspaceSymbolParams) ([]lsp.SymbolInformation, error) {
+	documentManager := service.MustGet[workspace.DocumentManager](p.sc)
+	query := strings.ToLower(params.Query)
+	maxCount := p.filter.MaxSymbolCount()
+
+	var symbols []lsp.SymbolInformation
```
msujew, line 73: Suggestion: Can we set an initial capacity for this, assuming `maxCount > 0`? That way, we never need to resize the slice.

test/fixture.go
```diff
@@ -307,3 +308,106 @@ func locationInRange(loc core.TextLocation, r core.TextRange) bool {
 	}
 	return true
 }
+
+// FindWorkspaceSymbol searches workspace symbols by name.
+// Returns the symbol and true if found, or nil and false if not found.
+func (f *Fixture) FindWorkspaceSymbol(name string) (*lsp.SymbolInformation, bool) {
```
msujew: Suggestion: Can we restructure this similar to the diagnostics? I.e. an `ExpectWorkspaceSymbols` function that can be chained like:

```go
doc.ExpectWorkspaceSymbols("some-search-string").
  Contains("a", "b", "c"). // <-- use variadic arrays for this
  SymbolKind("c", lsp.SymbolKind.Variable)
```

This would have the benefit (aside from better API) to only have to call the workspace symbol request once.

msujew (commented):
Thanks a lot! I think that it would make sense to reuse the fuzzy matching service implemented in the completion PR. That way, we don't have to maintain multiple fuzzy matching implementations.

ssmifi:
Thanks for the review. Also uses the fuzzy matcher service now. I moved my fuzzy tests over, but if you think they are too much, I can remove them.

internal/grammar/workspace_symbol_provider_test.go
```diff
@@ -0,0 +1,70 @@
...
+	fixture.ExpectWorkspaceSymbols("Person").
+		SymbolKind("Person", lsp.Interface).   // Interface
+		SymbolKind("PersonRule", lsp.Function) // Parser rule (also matches "Person")
+	fixture.ExpectWorkspaceSymbols("name").SymbolKind("name", lsp.Property) // Field
+
+	// Test 4: Case-insensitive matching
+	fixture.ExpectWorkspaceSymbols("person").ExactMatch("Person", "PersonRule")
+
+	// Test 5: No match
+	fixture.ExpectWorkspaceSymbols("XYZ").ExactMatch()
```
msujew: Suggestion: Can we run each of them as subtests via `t.Run()`? That gives us a better understanding in case of regressions.

msujew (approved):
Thanks, looks good to me. Just one minor improvements to the tests, see below.

PR #86, opened 2026-06-02
Support token groups

msujew:
Closes https://github.com/TypeFox/fastbelt/issues/49

Adds support for token groups like:

```
token group IdOrKeyword {
  ID
  // Matches any keyword with a letter within
  keywords /\w/
}
```

and others (see tests). In order for this to correctly interact with the parser/completion/error recovery, all of the respective runtimes+generators required a few changes. Also adds a dedicated testing language `internal/languages/token_groups` to provide extensive testing for the feature.

Adds validation checks for the new feature:
1. Ensures that token groups can't recursively call each other.
2. Ensure that token groups can't contain hidden or comment tokens.


spoenemann (commented):
Nice! Please add documentation to `grammar/doc.go`. In particular, the distinction between composites and token groups needs to be explained (when to use what).

I'm not sure about the regexp syntax. What does it mean exactly? Does it only apply to matching keywords?

msujew:
> I'm not sure about the regexp syntax. What does it mean exactly? Does it only apply to matching keywords?

Yes, only to keywords - the alternative would be to spell out each keyword individually. In languages where all letter-based-keywords are potential identifiers (such as in PL/I, among other languages), one would need to specify each one manually, leading to a lot of room for error (i.e. forgetting one), especially because PL/I features hundreds of keywords. The regexp approach tries to solve this problem by not having to specify each one manually.

examples/statemachine/lexer_gen.go
```diff
@@ -9,7 +9,7 @@ import (
 	"unicode/utf8"
 )
 
-const Keyword_EqualsGreaterThan_Idx = 0
+const Keyword_EqualsGreaterThan_Idx = 1
```
Lotes, line 12: Has Index 0 a special meaning here?

internal/generator/lexer_generator.go
```diff
@@ -23,36 +26,42 @@ type GenerateTokenTypesResult struct {
...
 		TokensCode:        make([]codegen.Node, len(tokens)),
-		TokenTypeNames:    make([]string, keywordsCount+len(tokens)),
-		TokenTypeVarNames: make([]string, keywordsCount+len(tokens)),
+		TokenGroupCode:    make([]codegen.Node, len(tokenGroups)),
+		TokenTypeNames:    make([]string, keywordsCount+len(tokens)+len(tokenGroups)),
+		TokenTypeVarNames: make([]string, keywordsCount+len(tokens)+len(tokenGroups)),
 		TokenTypeIds:      make(map[string]int),
 		Imports:           map[string]bool{},
 	}
+	// Starting with 1 - prevent clash with EOF (index 0)
```
Lotes, line 51: Aha! Confusing. Never saw this hint. It is not so obvious by reading the code. Can we generate EOF as well in the lexer code? It feels unnatural to start at 1.

Lotes: AND: If you take one away from the beginning, I would expect you to add +1 for teh right indexing somewhere. Have not seen it yet.

internal/grammar/completion_parser_gen.go
```diff
@@ -94,74 +99,66 @@ func (p *CompletionParser) ParseGrammar() {
...
 			p.cp.MarkAssignment("Fields")
-			p.state.EnterRule(78)
+			p.state.EnterRule(82)
 			p.ParseField()
 			p.state.ExitRule()
 			p.cp.ClearAssignment()
-			p.cp.RecordSnapshot(79)
-			p.state.Sync(79)
+			p.cp.RecordSnapshot(83)
+			p.state.Sync(83)
```
Lotes: 🪝 🏴‍☠️ garrr these ugly magic numberrrs!

internal/grammar/validator.go
```diff
@@ -547,3 +554,63 @@ func doInterfaceIsAssignableTo(source Interface, target Interface, visited map[s
...
+
+func checkInvalidTokensInGroup(tg TokenGroup, accept core.ValidationAcceptor) {
+	for _, ext := range tg.TokenRefs() {
+		abstractToken := ext.Ref(context.Background())
+		if abstractToken == nil {
+			continue
+		}
+		if token, ok := abstractToken.(Token); ok {
+			tokenType := token.Type()
+			special := tokenType == "hidden" || tokenType == "comment"
```
Lotes, line 604: Is `tokenType` really the right term here?
`channel`, `mode`, `visibleAs`, `kind`?

internal/grammar/validator_test.go
```diff
@@ -291,3 +321,50 @@ func TestKeywordAssignedToNonStringField(t *testing.T) {
...
+		token group <|Y|> { X }
+	` + commonTokens)
+	for _, label := range []string{"X", "Y"} {
+		diag := doc.ExpectDiagnostic(label)
+		diag.WithSeverity(core.SeverityError)
+		diag.WithCode(ValidateRecursiveTokenGroup)
+	}
+}
+
+func TestTokenGroupRecursiveNegative(t *testing.T) {
```
Lotes, line 353: Why is it negative?

internal/grammar/validator_test.go
```diff
@@ -291,3 +321,50 @@ func TestKeywordAssignedToNonStringField(t *testing.T) {
...
+	doc.AssertNoErrors()
+}
+
+func TestTokenGroupWithInvalidToken(t *testing.T) {
+	f := test.New(t, CreateServices())
+	doc := f.Parse(`
+		grammar Test;
+		token group X { <|WS|> }
+	` + commonTokens)
+	diag := doc.ExpectDiagnostic("WS")
```
Lotes, line 368: May add a comment, why it is invalid.

internal/languages/completion/completion.fb
```diff
@@ -74,17 +84,33 @@ interface J extends Obj {
...
 }
 
+// two assignments to the same target type - dedup check
 K: "k" (Ref1=[Declare:ID] "x" | Ref2=[Declare:ID] "y");
 
-
+// fully optional group in front of mandatory content
 L returns Obj: "l" ("optional" "and")? "then" "end";
 
+token group SomeTokenGroup {
```
Lotes, line 101: Question: Does a different order of token groups influence the parsing/lexing?

And inside the group?

util/collections/bitset.go
```diff
@@ -0,0 +1,143 @@
...
+package fastbelt
+
+// BitSet is a simple implementation of a bitset that supports insertion,
+// deletion, and membership testing. It is designed to be memory efficient for
+// sparse sets of integers, where the integers can be large but the number of
+// integers in the set is relatively small.
+//
+// Inside of Fastbelt, BitSets are mostly used to handle sets of token types
+// in the parser engine.
+type BitSet struct {
```
Lotes, line 14: Don't we have this already?
/internal/automatons/bitmask.go

Lotes: I use strings there, but bits would be cooler actually. Maybe a follow up PR if applicable...

internal/languages/completion/completion.fb
msujew: No difference at all. The lexer generator performs topological sorting based on which token group includes what other token group - inside the group, there is only the `BitSet` which has no specific ordering.

internal/generator/lexer_generator.go
msujew: The [EOF token type](https://github.com/TypeFox/fastbelt/blob/42c90bcd199530ca504cbea2292a45c5932a8715/token.go#L88-L102) already exists, not a good idea to generate it - it would make it impossible to use `EOF` in framework code.

internal/grammar/validator.go
msujew: It's not ideal, but it's already in use - I would change this in a separate PR.

spoenemann:
>> I'm not sure about the regexp syntax. What does it mean exactly? Does it only apply to matching keywords?
>
> Yes, only to keywords - the alternative would be to spell out each keyword individually. In languages where all letter-based-keywords are potential identifiers (such as in PL/I, among other languages), one would need to specify each one manually, leading to a lot of room for error (i.e. forgetting one), especially because PL/I features hundreds of keywords. The regexp approach tries to solve this problem by not having to specify each one manually.

That makes sense, but I'm concerned about the grammar syntax:

```
token group IdOrKeyword {
  ID
  /\w/
}
```
This looks strange because "token group" sounds like the regexp here is a token defined inline – but it's shadowed by the ID token. Could we make the semantics more explicit?
```
token group IdOrKeyword {
  ID
  keywords /\w/
}
```

spoenemann:
BTW shouldn't the regexp be rather `/\w+/`? I'd find it clearer to write it such that it matches the full keyword, not just a substring.

msujew:
> BTW shouldn't the regexp be rather /\w+/? I'd find it clearer to write it such that it matches the full keyword, not just a substring.

That's up to the adopter. They could be very specific if they want to, with `/^\w+$/` as well.

> This looks strange because "token group" sounds like the regexp here is a token defined inline – but it's shadowed by the ID token. Could we make the semantics more explicit?

That's a good point. I'll think of something.

Edit: Couldn't think of something, used the `keywords Regexp` approach.

internal/grammar/grammar.fb
```diff
@@ -67,19 +74,32 @@ interface AbstractRuleWithBody extends AbstractRule {
...
 Token: (Type="hidden" | Type="comment")? "token" Name=ID ":" Regexp=RegexLiteral ";";
 
+interface TokenGroup extends AbstractTokenRule {
+    TokenRefs []*AbstractTokenRule
+    Regexps []string
+    Keywords []Keyword
+}
+
+TokenGroup: "token" "group" Name=ID "{"
+    (TokenRefs+=[AbstractTokenRule:ID] | "keywords" Regexps+=RegexLiteral | Keywords+=Keyword)*
```
spoenemann, line 100: The new `keywords` keyword should be added to fastbelt.tmLanguage.json

util/collections/bitset.go
spoenemann, line 14: Shall we move this to `util/collections`?

spoenemann (approved):
👍

PR #87, opened 2026-06-02
Support document highlight LSP feature

msujew:
Does as the title says. Works just like the references provider - but just for the current file.

Slightly adjusts the `ReferencesFinder` to have a `TargetURI` property. Implements a fast path if that property is set to only look at the selected URI.


spoenemann (approved):
👍

PR #88, opened 2026-06-03
Fix automatons determinize algorithm

Lotes:
There were some edge cases were determinize is not working properly.
Complement and Intersect can only work on DFAs. Especially when the dead state is gone, but needed, these two algorithms can fail.


internal/automatons/construct.go
```diff
@@ -32,8 +32,7 @@ func (ck *ConstructionKit) Consume(characters *RuneSet) *NFA {
 	builder.SetStartState(start)
 	builder.AcceptState(end)
 	builder.AddTransitionForRuneSet(start, end, characters)
-
-	return builder.Build()
+	return ck.finalize(builder)
```
Lotes, line 35: Might be too powerful, just insert the dead end instead.

msujew (commented):
Did you run `go generate ./...` on this? I would've thought that this changes the generated lexer files.

internal/automatons/construct.go
Lotes: Now I think it is good enough. The desired DFA is the following one.

```mermaid
flowchart LR
    1 --a--> 2
    1 --^a--> 3
    3 --*--> 3
    2 --*--> 3
```

But "a" could be empty or full, which complicates the algorithm a bit.

msujew (commented):
This seems to have a major impact on performance. It reduces the `BenchmarkLexer` from ~180MB/s to ~1MB/s on my machine.

Debugging through it, it seems like rules like `ID` always read until the end of the input, even when encountering a character that does not fit their regexp. Something is very wrong there.

internal/regexp/regexp_generator.go
```diff
@@ -14,10 +14,13 @@ type GenerateTransitionsUsingBinarySearchResult struct {
 	Code   codegen.Node
 }
 
-func GenerateTransitionsUsingBinarySearch(bySource *automatons.RuneRangeTargetsMapping, source int, tokenName string, imports map[string]bool) GenerateTransitionsUsingBinarySearchResult {
+func GenerateTransitionsUsingBinarySearch(bySource *automatons.RuneRangeTargetsMapping, source int, tokenName string, imports map[string]bool, isAcceptanceReachable map[int]bool) GenerateTransitionsUsingBinarySearchResult {
 	transitions := make([]automatons.RuneRangeMappingSection[automatons.Targets], 0)
-	if bySource != nil {
+	if bySource != nil && isAcceptanceReachable[source] {
```
Lotes: Instead of asking for a dead end during execution. I am avoiding the generation, which makes the automaton smaller and faster again.

examples/statemachine/lexer_gen.go
```diff
@@ -263,14 +278,16 @@ var Token_ID = core.NewTokenType(
...
 	{0x0000005A00000041, 0x0000005F0000005F, 0x0000007A00000061},
+	{},
 	{0x0000003900000030, 0x0000005A00000041, 0x0000005F0000005F, 0x0000007A00000061},
 }
 var Token_ID_Next = [][]int{
-	{1, 1, 1},
-	{1, 1, 1, 1},
+	{2, 2, 2},
+	{},
+	{2, 2, 2, 2},
```
msujew: Suggestion: Looking at this - there is no transition which is able to enter the dead state. Can we somehow completely eliminate it, if it's not reachable?

internal/automatons/nfa.go
```diff
@@ -34,3 +34,68 @@ func (nfa NFA) String() string {
 	}
 	return result
 }
+
+func (nfa NFA) DotFile() string {
+	result := "digraph NFA {\n"
+	result += "  rankdir=LR;\n"
+	result += "  node [shape=circle];\n"
```
msujew, line 48: Minor: Seems unused?

msujew (commented):
Thanks, much better!

internal/automatons/nfa.go
Lotes: It is meant as a helper for debugging.

Lotes: Bad?

examples/statemachine/lexer_gen.go
```diff
@@ -474,18 +459,16 @@ var Token_ML_COMMENT_Lookup = [][]int64{
 	{0x0000002A0000002A},
 	{0x0000002900000000, 0x0000002A0000002A, 0x0010FFFF0000002B},
 	{0x0000002900000000, 0x0000002A0000002A, 0x0000002E0000002B, 0x0000002F0000002F, 0x0010FFFF00000030},
-	{0x0000002900000000, 0x0000002A0000002A, 0x0010FFFF0000002B},
-	{0x0000002900000000, 0x0000002A0000002A, 0x0010FFFF0000002B},
+	{},
```
Lotes, line 462: There are still these cases... wondering where this is coming from.
Maybe a forgotten Minimize? ^^* very weird...

Lotes: Space for optimization...

Lotes: Kommando zurück: This is mostly the accepting state. All good.

internal/automatons/nfa.go
msujew: Not necessarily bad, but I don't feel comfortable having semi-dead code in the repo. Move this into a test file + a test case for this? That way you can still debug it easily.

Lotes: Done.

internal/automatons/nfa_test.go
```diff
@@ -0,0 +1,102 @@
...
+	return nfa
+}
+
+func TestComputeAcceptanceReachability_ConstructedDFA(t *testing.T) {
+	aOrAB := makeDFA()
+
+	actual := aOrAB.ComputeAcceptanceReachability()
+	expected := map[int]bool{
+		0: true,
+		//1: false, //unreachable, there is only one because the consstruction kit makes DFA-like NFAs which have only one dead state
```
msujew: Minor: There's no need to comment out the `1: false` if you want to be explicit.

msujew (approved):
Thanks, looks good to me 👍

internal/automatons/nfa_test.go
Lotes: It is a difference. There is no entry `1: false` in actual. Maybe I should make setting false explicit in the algorithm...

PR #89, opened 2026-06-03
Package documentation: lexer, textdoc, util/*

spoenemann:
Adding more documentation:

- lexer package
- textdoc package
- All util packages
- Removed redundancy between root package doc and README.md, which is included in the public documentation

### DIFF
--- a/lexer/doc.go
+++ b/lexer/doc.go
@@ -38,7 +38,9 @@
 //     types in a fixed-size map keyed by rune % 256 (each [core.TokenType]'s
 //     StartChars slice).
 //  2. Run each candidate's [core.Matcher] and keep the longest match (maximal
-//     munch). Keywords and regex token rules compete on equal footing.
+//     munch). Among equal-length matches, the first registered token type wins;
+//     generated lexers list keywords before regex token rules, so keywords take
+//     precedence when both match the same span.
 //  3. Route the match by [core.TokenType.Group]: default tokens go to
 //     [LexerResult.Tokens], hidden tokens are dropped, comments go to
 //     [LexerResult.Comments], and other groups are collected in
--- a/lexer/doc.go
+++ b/lexer/doc.go
@@ -1,0 +1,56 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+// Package lexer provides the shared lexer runtime used by Fastbelt-generated
+// languages. It does not implement lexing for any particular language; each
+// grammar gets its own generated lexer that wires this package together with
+// language-specific [core.TokenType] values.
+//
+// Run [typefox.dev/fastbelt/cmd/fastbelt] generate on a .fb grammar file to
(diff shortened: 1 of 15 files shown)


lexer/doc.go
```diff
@@ -0,0 +1,56 @@
...
+//
+// # Lexing model
+//
+// Fastbelt uses a table-driven scanner with the following approach:
+//
+//  1. At each input offset, decode the current rune and look up candidate token
+//     types in a fixed-size map keyed by rune % 256 (each [core.TokenType]'s
+//     StartChars slice).
+//  2. Run each candidate's [core.Matcher] and keep the longest match (maximal
+//     munch). Keywords and regex token rules compete on equal footing.
```
msujew: What is "maximal munch"? :D

Also note that Keywords are actually preferred to token rules, if they lex the same input, since they have higher priority (appear first in the list).

lexer/lexer.go
```diff
@@ -152,6 +165,9 @@ func (l *DefaultLexer) Lex(input string) *LexerResult {
 
 const maxChar = 256
 
+// NewDefaultLexer returns a lexer that recognizes the given token types.
+// The order of arguments is not significant; matching uses longest match at
+// each position.
```
msujew: This is incorrect. Even though it matches on longest match - the first token encountered with the longest match wins (otherwise we run into issues with keywords, which would be unmatchable).

util/service/doc.go
```diff
@@ -0,0 +1,68 @@
...
+//
+//	func CreateServices() *service.Container {
+//	    sc := service.NewContainer()
+//	    SetupServices(sc)
+//	    sc.Seal()
+//	    return sc
+//	}
+//
+// Framework code that must resolve a dependency at runtime calls
+// [MustGet] after the container has been sealed.
```
msujew: Feels superfluous? How else should it get the dependency?

util/service/doc.go
```diff
@@ -0,0 +1,68 @@
...
+// registers its default implementations into a container:
+//
+//   - [typefox.dev/fastbelt/textdoc.SetupDefaultServices]
+//   - [typefox.dev/fastbelt/linking.SetupDefaultServices]
+//   - [typefox.dev/fastbelt/workspace.SetupDefaultServices]
+//   - [typefox.dev/fastbelt/server.SetupDefaultServices]
+//
+// These functions are idempotent: they call [Has] before [Put] and skip types
+// that are already registered, so a language can pre-register custom
+// implementations and still call the framework defaults for everything else.
```
msujew: Suggestion: Maybe add that `Override` can be used to register custom implementations _after_ calling these setup functions? As an alternative to the pre-registration process.

msujew (commented):
Thanks, looks quite good 👍

lexer/doc.go
spoenemann: 👉  https://en.wikipedia.org/wiki/Maximal_munch

I adjusted the wording of the second sentence to:

> Among equal-length matches, the first registered token type wins; generated lexers list keywords before regex token rules, so keywords take precedence when both match the same span.

msujew (approved):

PR #91, opened 2026-06-04
WASM Prototype

msujew:
This provides the glue required for Fastbelt to work correctly in a WASM (web) environment. Adds a prototype for the statemachine language. Can be tested by starting the `Statemachine Web Extension` launch config.

We can later think about extracting the JS-glue into its own npm package like `@fastbelt/wasm` or something.

In my initial benchmarks, the WASM runtime is roughly 3x slower than the single-threaded desktop binary - I think this is quite good.

### DIFF
--- a/internal/vscode-extensions/statemachine/src/extension.web.ts
+++ b/internal/vscode-extensions/statemachine/src/extension.web.ts
@@ -1,0 +1,61 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+import * as vscode from 'vscode';
+import type { LanguageClientOptions } from 'vscode-languageclient/browser.js';
+import { LanguageClient } from 'vscode-languageclient/browser.js';
+
+let client: LanguageClient;
+
+// This function is called when the extension is activated in a browser host
+// (e.g. VS Code Web). The language server runs as a WebAssembly module inside a
+// Web Worker instead of as a spawned process.
+export async function activate(context: vscode.ExtensionContext): Promise<void> {
+    client = await startLanguageClient(context);
+}
+
+// This function is called when the extension is deactivated.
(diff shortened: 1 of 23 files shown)


internal/vscode-extensions/statemachine/src/extension.web.ts
```diff
@@ -0,0 +1,61 @@
...
+    // The worker hosts the WASM-compiled Go language server and bridges
+    // JSON-RPC messages to it over postMessage. We read the WASM bytes here (the
+    // extension host can access its own bundled resources regardless of scheme)
+    // and transfer them to the worker as its first message, because the worker
+    // cannot fetch/resolve resources from its own blob/file URL.
+    const workerUri = vscode.Uri.joinPath(context.extensionUri, 'dist', 'server.worker.js');
+    const wasmUri = vscode.Uri.joinPath(context.extensionUri, 'dist', 'server.wasm');
+    const worker = new Worker(workerUri.toString(true));
+    const wasm = await loadWasm(wasmUri);
+    worker.postMessage({ __init: true, wasm }, [wasm.buffer]);
```
sailingKieler, line 36: For curiosity: Why only transferring ownership of `wasm.buffer` instead of entire `wasm`?

sailingKieler (approved):
This is sophisticated stuff, took quite some time for me to understand it.
The most questions I had were answered by the (/some) comments, but you need to know what to look for.

It took me a while to realize that the Go runtime brings
`GOROOT/lib/wasm/wasm_exec.js`

LGTM 👍

internal/vscode-extensions/statemachine/src/extension.web.ts
msujew: `wasm` itself (as an `Uint8Array`) isn't a `Transferable`. Only the underlying buffer is actually transferable to the worker thread. Trying to do so generates an error:

> Failed to execute 'postMessage' on 'Worker': Value at index 0 does not have a transferable type..

PR #92, opened 2026-06-04
Added logo

spoenemann:


internal/vscode-extensions/fastbelt/package.json
```diff
@@ -10,8 +10,9 @@
     "Programming Languages"
   ],
   "license": "MIT",
+  "icon": "data/fastbelt-logo.png",
```
msujew, line 13: Can we also add this as a file icon? See [langium](https://github.com/eclipse-langium/langium/blob/c28c0c4339ace747f5208d7f227d43cb7eb2ca7b/packages/langium-vscode/package.json#L50-L53)

README.md
```diff
@@ -1,4 +1,4 @@
-# Fastbelt
+![Fastbelt](https://raw.githubusercontent.com/TypeFox/fastbelt/refs/heads/main/fastbelt-full-logo.png)
```
msujew: Question: Is the full URL necessary for the go package registry?

msujew (commented):
Thanks 👍

README.md
spoenemann: Yes it doesn't copy images.

spoenemann: Oh actually relative should work, the registry rewrites such URLs when the repo is on GitHub.

msujew (approved):
Great, thank you!

PR #93, opened 2026-06-04
Properly compute UTF16 positions

msujew:
Fixes an issue I've noticed when accidentally typing in non-ascii characters into the editor: The `Overlay` kept outdated text in there due to incorrect offsets. By default, the LSP will assume that all range columns are computed in UTF16 (mainly because the reference implementation is in JavaScript, which uses UTF16 for all strings). We should align to this default (we could also try to communicate that we use UTF8, but not all clients accept this) in our LSP position handling code.

This updates two main parts which are related to this:
1. The lexer now computes column information using UTF16 (which slightly slows down lexing)
2. The `textdoc` infrastructure can now compute byte offsets <-> UTF16 position. Previously, all positions were handled as byte-positions (not even UTF8!).

In addition to the automated tests, can also be tested using the statemachine extension using a comment like `/* 😀 */`. All lsp requests should behave as expected, even in the presence of this comment.

### DIFF
--- a/examples/statemachine/statemachine_test.go
+++ b/examples/statemachine/statemachine_test.go
@@ -221,6 +221,41 @@ func TestFindNodeAtLabel(t *testing.T) {
 	assert.Equal(t, "S", sm.Name())
 }
 
+func TestTokenColumnUTF16(t *testing.T) {
+	// Minimal single-line statemachine used as input for column checks.
+	const suffix = ` statemachine S events e initialState idle state idle e => idle end`
+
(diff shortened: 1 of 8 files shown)


spoenemann (commented):
Looks good, but I think we should explicitly state our assumed encoding in the server capabilities:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocuments

spoenemann (approved):

PR #98, opened 2026-06-08
added generation of type specific 'MarshalJSON' & 'UnmarshalJSON' implementations

sailingKieler:
Contributes support for exporting and importing ASTs in Json.
Includes:
* a generator component producing `json_gen.go` files
  * they contribute methods turning AstNodes into impls of  `json.Marshaler` & `json.Unmarshaler`
* generic `MarshalJSON` & `UnmarshalJSON` implementations for `Reference`
* `util/json/json.go` demonstrates the procedure of importing an AST in json into a Fastbelt container

* a simple round trip AST export and import test including cross references to other documents (circular) by means of the Arithmetics test language,
  * focus is on successfully resolving the (cross document) cross-references
* a roundtrip AST export and import test (no external cross-refs by now) by means of the Grammar language,
  * focus is on completeness of feature processing and cardinalities:
    `boolean`, `string`, `AstNode`, `Reference` x `1`, `?`, `*`, `+`

Note:
* Minor supplements in `fixture.go`
* Fix of a potential `nil` pointer de-reference in `symbol.go`

There's one known potential issue:
During the import of an AST in json the `Token` instances are always initialized with token type `Token_ID`.
The leads to a compiler issue if there is no `ID` token type.

We could improve this by an analysis of the Parser rules, but that will never be complete.
Imagine the grammar defines interfaces of types that are never instantiated by parser rules, but the data are always imported vom Json snapshots.
To mitigate this we might offer an opportunity in the grammar to determine a token type for fields of type `string` or `composite`. To be discussed.


reference.go
```diff
@@ -292,3 +294,82 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+// MarshalJSON serializes the reference as a JSON object containing the URI of the resolved target's document.
+// With this function Reference[T] implements json.Marshaler.MarshalJSON() and the method is called by `json.Marshal()` and `json.MarshalIndent()`.
+func (r *Reference[T]) MarshalJSON() ([]byte, error) {
+	// A nil reference marshals as JSON null.
+	if r == nil {
+		return []byte("null"), nil
+	}
+	// We expect at this point that all references have been attempted to resolve.
+	if !r.resolved.Load() {
+		return nil, errors.New("reference not resolved")
```
spoenemann: Proposal: generate a partial reference like in Langium:

https://github.com/eclipse-langium/langium/blob/7ad7c76e85134185146e1a99bd352e7527ed1fec/packages/langium/src/serializer/json-serializer.ts#L174-L177

reference.go
```diff
@@ -292,3 +294,82 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+		uri = doc.URI.WithFragment("todo-fragment").StringUnencoded()
+	} else {
+		return nil, errors.New("unexpected state: resolved reference has no document")
+	}
+
+	return json.Marshal(struct {
+		RefText string `json:"refText"`
+		URI     string `json:"uri,omitempty"`
+		Err     string `json:"error,omitempty"`
+	}{
```
spoenemann, line 345: This struct def should be exported. And I propose aligning this with Langium for cross-framework compatibility:

https://github.com/eclipse-langium/langium/blob/7ad7c76e85134185146e1a99bd352e7527ed1fec/packages/langium/src/serializer/json-serializer.ts#L96-L105

reference.go
```diff
@@ -292,3 +294,82 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+		RefText: r.unit.String(),
+		URI:     uri,
```
spoenemann: In Langium, we omit the URI if the reference target is in the same document. This can reduce the size of generated JSON by a lot.

Thought: can we include something like a URI converter here as well? Exporting JSONs with absolute file URIs would be pretty useless in many cases.

https://github.com/eclipse-langium/langium/blob/7ad7c76e85134185146e1a99bd352e7527ed1fec/packages/langium/src/serializer/json-serializer.ts#L161-L167

sailingKieler: > In Langium, we omit the URI if the reference target is in the same document. This can reduce the size of generated JSON by a lot.

That's achieved by the `omitempty` setting. Of course, the prop value shouldn't be set upfront.

> Thought: can we include something like a URI converter here as well? Exporting JSONs with absolute file URIs would be pretty useless in many cases.

Yes, that's my plan.

reference.go
sailingKieler: > This struct def should be exported.

Why?
It's purely technical and kind of a vehicle.
It might make sense to rely in the two (Un)Marshal methods on the same definition, though.

>And I propose aligning this with Langium for cross-framework compatibility:

That's my plain, too.

reference.go
sailingKieler: This is just preliminary, I don't know which `context` instance to rely on for calling `Ref(...)`, yet
(since this function is called by `json.Mashall(...)`, not by ourselves), and whether we should attempt to resolve here, at all.

AFAICS, `r.resolved` is `true` if resolution has been attempted, regardless of success or failure.
Hence, this guards stops the serialization, e.g., if the document hasn't been build until `DocStateLinked` at least.

sailingKieler (commented):
Thanks for the hints Miro.
This state is a very early one, it doesn't represent the one I have in mind already.

I also seek for a solution that is compatible with Langium, s.t. we could use it as an interface for data exchange

More changes are coming soon :-)

Lotes:
Just an idea: We are only generating a JSON here, right?
If someone wants to use the output, he needs to model the meta model, first. Or he hopes everything is like in his imagination.
A lot of languages understand JSON Schema and can generate type information from the schema. Then, deserializing and validating the output should be very simple.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,299 @@
...
+import (
+	"strings"
+
+	"typefox.dev/fastbelt/internal/grammar"
+	"typefox.dev/fastbelt/util/codegen"
+)
+
+func GenerateJSON(grammar grammar.Grammar, packageName string) string {
+	node := codegen.NewNode()
+	node.AppendLine("// Code generated by typefox.dev/fastbelt/cmd/fastbelt. DO NOT EDIT.")
```
Lotes, line 16: We could move this line to a constant that is shared over all generators.

examples/arithmetics/json_test.go
```diff
@@ -0,0 +1,415 @@
...
+	cleanup := func() {
+		f.Clear()
+	}
+
+	language := service.MustGet[workspace.LanguageID](f.Services())
+	documents := service.MustGet[workspace.DocumentManager](f.Services())
+	builder := service.MustGet[workspace.Builder](f.Services())
+
+	t.Run("selfContained", func(t *testing.T) {
+		t.Cleanup(cleanup)
```
Lotes, line 40: In my eyes it looks a bit weird that you clean up before doing something.

`defer t.Cleanup(cleanup)` would clean up after the test was processed.

examples/arithmetics/json_test.go
```diff
@@ -0,0 +1,415 @@
...
+			case "/testB.json":
+				assert.Equal(t, circularBJsonAdj, string(res))
+			default:
+				assert.Failf(t, "Unexpected document %s", path)
+			}
+		}
+	})
+}
+
+const selfContainedJson = `{
```
Lotes, line 191: We could use [embedded files](https://pkg.go.dev/embed) here.
Then the "code space" gets not polluted with big escaped JSON strings.
But I am also fine like it is now. Just wanted to show an opportunity to save > 200 LOC that are actually a different language.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,299 @@
...
+
+	node.AppendLine("func (i *", iface.Name(), "Impl) MarshalJSON() ([]byte, error) {")
+	node.Indent(func(n codegen.Node) {
+		// preprocess []string fields: they are stored as []*core.Token internally but shall be serialized as plain strings;
+		// instead of implementing 'MarshalJSON()' for *core.Token, a preprocessing loop calling '.String()' for each item
+		// is added for each of such fields
+		var stringListFields = map[string]string{}
+		for _, field := range fields {
+			if field.Array && (field.GType == TOKEN_TYPE || field.GType == COMPOSITE_TYPE) {
+				varName := strings.ToLower(field.Name[:1]) + field.Name[1:]
```
Lotes: Move out repeating snippet like name concatenation to a function.
Or may collect frequently used variable names, instead of computing them in place.
For: `varName`, `jsonTag`.
Also because you do not have to change `N` places if you decide that the names shall change.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,299 @@
...
+
+func genReturnErr(node codegen.Node) {
+	node.Indent(func(n2 codegen.Node) {
+		n2.AppendLine("return err")
+	})
+	node.AppendLine("}")
+}
+
+func generateDispatchingUnmarshalFunc(node codegen.Node, g grammar.Grammar) {
+	node.AppendLine(`// Unmarshal decodes data into an instance of type T by reading the "$type" field,`)
```
Lotes, line 244: This function outputs a string that differs only in the grammar name.
I am wondering if we can generate this part in a less verbose way.
Actually it is like a utility function that could almost live in a separated file besides the generated functions. Can we move it out the generated folder? The advantage would be that you get displayed syntax errors for free.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,299 @@
...
+	node.AppendLine(")")
+	node.AppendLine()
+
+	generateFunctions(grammar, node)
+
+	return FormatIfPossible(node.String())
+}
+
+func generateFunctions(grammar grammar.Grammar, node codegen.Node) {
+	node.AppendLine(("func newToken(tokenType *core.TokenType, view string) *core.Token {"))
```
Lotes, line 38: This utility function does not differ from other generations. It would make sense to move it outside the generated folder.

internal/grammar/json_test.go
```diff
@@ -0,0 +1,264 @@
...
+		path string
+	}{
+		{"Fastbelt Grammar language", "grammar.fb"},
+		{"Completion test language", "../languages/completion/completion.fb"},
+		{"Lookahead test language", "../languages/lookahead/lookahead.fb"},
+		{"TokenGroups test language", "../languages/token_groups/token_groups.fb"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			grammmar, err := os.ReadFile(tt.path)
```
Lotes, line 34: Think of using the embed trick from above?

internal/grammar/json_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+	return doc1.Root(), doc2.Root
+}
+
+var REF_TOKEN_TYPE = reflect.TypeFor[*core.Token]()
+
+// assertEqualAst recursively compares two AST subtrees node by node. Diffs are
+// reported via t.Errorf with the node's document path from core.PathOf.
+// Returns true when at least one difference was detected.
+func assertEqualAst(t testing.TB, expected, actual core.AstNode) bool {
```
Lotes: It looks cleaner if these helper function would live in a own file.
But I think it is ok FTM, as long as there is only one test file that uses them.

reference.go
```diff
@@ -292,3 +295,147 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+	})
+}
+
+// UnmarshalJSON revives the reference as a JSON object. It inspects properties named 'ref', 'refText', and 'err'.
+// 'ref' is expected to contain a URI string denoting the target node within the same or another document, may be absent if the reference was unresolvable before serializing.
+// 'refText' denotes the cross reference string.
+// 'err' contains the error msg if the reference was unresolvable before serializing
+// With this function Reference[T] implements json.Unmarshaler.UnmarshalJSON() and the method is called by `json.Unmarshal()`.
+func (r *Reference[T]) UnmarshalJSON(data []byte) error {
+	aux := &struct {
```
Lotes, line 358: You should make a private type here, because it is repeating.

Lotes (commented):
Here are some comments. In the reference part I was not so deep.
I hope it helps :-) ...
About the mentioned "embed trick", I am not sure whether there are downsides. Using it within the tests should be fine.

examples/arithmetics/json_test.go
sailingKieler: From the docs of `*common.Cleanup(...)`: 
`Cleanup registers a function to be called when the test (or subtest) and all its subtests complete.`

Thus, it is supposed to do what we expect, i.e., do clean up once the test is completed.

examples/arithmetics/json_test.go
sailingKieler: As discussed in the meeting, it won't be super beneficial here but should keep that opportunity in mind for more complex tests.

internal/generator/json_generator.go
sailingKieler: Indeed, it is basically a utility function, and I had it in the framework first.
However, it appeared fairly orphaned there with almost no context, so I added it to the generated code, also because it's called in the generated code.
This way it will always be consistent with the generated code.
It's a matter of taste.

internal/generator/json_generator.go
sailingKieler: It also a utility that is tightly coupled to its callers that we might change depending on the open question regarding the token type, therefore it's non-exported.
 In the face of #128 we might even drop it and inline its impl.

reference.go
sailingKieler: That private type will be visible in the entire Fastbelt core package, which doesn't make any sense.
Yes we could chose a name the makes misuses obvious but, honestly, I don't see any benefit in that.

internal/grammar/json_test.go
sailingKieler: Moved it into a separate file named `test/doc_fixture_util.go`, as the function seems to be too long & complex to add it to one of the existing files in `test`.

internal/grammar/json_test.go
sailingKieler: I though about this for a moment, but it doesn't really make sense to me because of:
* this is a test, not production to be bundled into a self-contained executable.
* I would have to create a string var for each file to embed.
* I want the execution to fail with an ordinary test failure instead of _some_ compiler error if any file is absent/any path is wrong

internal/generator/json_generator.go
sailingKieler: Actually, this is AI generated code.
In case we need to update it we wanna prompt a coding agent to update it consistently, don't we? 😉

internal/generator/json_generator.go
sailingKieler: Sourced `strings.ToLower(field.Name[:1]) + field.Name[1:]` out to the composition of the `FieldInfo` description data in `type_generator.go`.

util/json/json.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+package fastbelt
```
spoenemann, line 5: The package name is inconsistent with the containing folder name. In addition, `json` is not a good package name because it conflicts with the standard library package. As this file currently exports only one function, I suggest moving it to the root of this module.

util/json/json.go
```diff
@@ -0,0 +1,61 @@
...
+
+	core "typefox.dev/fastbelt"
+	"typefox.dev/fastbelt/util/service"
+	"typefox.dev/fastbelt/workspace"
+)
+
+// UnmarshalAndBuildDocument uses the "encoding/json" entry point to unmarshal rootNode based on the given data string and builds the document.
+// Similar to parsing-based document loading the ast build up first including the reference, while the resolution of the references is done during the linking phase of the building process.
+// For properly linking references to other documents, a helper object is attached to the context given to builder providing access to other documents.
+func UnmarshalAndBuildDocument[T core.AstNode](sc *service.Container, document *core.Document, rootNode T, data []byte, ctx context.Context) error {
```
spoenemann, line 19: The ctx parameter should be the first per `context` package convention.

Wouldn't it be more useful to split the "Unmarshal" and the "Build" part into separate utilities? The former could go to the fastbelt (root) package, the latter to the workspace package.

symbols.go
```diff
@@ -125,6 +125,12 @@ func (c *mergedSymbolContainer) All() SymbolSeq {
 
 func (c *mergedSymbolContainer) ForType(targetType reflect.Type) SymbolSeq {
 	return extiter.FlatMap(c.containers, func(container SymbolContainer) SymbolSeq {
+		if container == nil {
+			// in case a document in the document manager hasn't been build yet or has been reset
+			// and is not included in the current build cycle its exported symbols container may be absent;
+			// map 'nil' to an empty seq in such cases
+			return func(yield func(*SymbolDescription) bool) {}
```
spoenemann, line 132: We have `EmptySymbolDescriptions` for this. BTW shall we rename that to `EmptySymbolSeq` to be consistent with the `SymbolSeq` type?

reference.go
```diff
@@ -292,3 +295,147 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+		owner:   r.owner,
+	}
+	r.getter = NewJsonReferenceGetter[T](aux.Ref)
+	if aux.Err != "" {
+		r.err = NewReferenceError(aux.Err)
+	}
+	return nil
+}
+
+type JsonLinkingHelper interface {
```
spoenemann, line 377: I'd consider making all these helper definitions private, including the NewJsonReferenceGetter function below. We can still make them public later if we see legitimate use cases.

In any case, please add some documentation to explain how these things work together.

spoenemann (commented):
Great! I like the overall approach.

internal/generator/json_generator.go
spoenemann: I'm in favor of moving this function to the framework by expecting the syntheticFactories map as an argument.

sailingKieler:
@spoenemann, @msujew One thing I'm still undecided about:
We could add the interfaces `[encoding/]json.Marshaler` and `[encoding/]json.Unmarshaler` to the generated AstNode interfaces, i.e. here:
https://github.com/TypeFox/fastbelt/blob/b789ac21086e7cf16077136254ca79865c68389b/internal/generator/type_generator.go#L176-L178

Then we can call `MarshalJSON()` and `UnmarshalJSON()` directly without casting or using the reflective API.
However, formatting is only provided by `json.MarshalIndent(...)`, and for type dispatching while reading and unmarshaling without a pre-created instance we need to use the (currently generated) generic `Unmarshal[...](...)` (maybe `UnmarshalAstNode`?) function.
Therefore I didn't add the interfaces in `type_generator.go`, yet.
What do you think?

symbols.go
sailingKieler: > BTW shall we rename that to EmptySymbolSeq to be consistent with the SymbolSeq type?

Makes sense IMHO.

reference.go
sailingKieler: > including the NewJsonReferenceGetter function below.

Makes sense, the uppercase initial is probably a leftover.

util/json/json.go
sailingKieler: Actually, my intention is to give a blue print how to do unmarshaling with cross-ref resolution.
We could move it into the docs, too, but then it's not compile checked anymore.

Therefore, I was hesitant to put it into the root package and pollute the namespace with these nice-to-have parts.
I'm fine with putting into `util`.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,295 @@
...
+			} else if field.Boolean {
+				n.AppendLine("if aux.", field.Name, "{")
+				n.Indent(func(n2 codegen.Node) {
+					n2.AppendLine("i.Set", field.Name, "(newToken(Token_ID, \"\"))")
+				})
+				n.AppendLine("}")
+			} else if field.GType == TOKEN_TYPE {
+				n.AppendLine("i.Set", field.Name, "(newToken(Token_ID, aux.", field.Name, "))")
+			} else if field.GType == COMPOSITE_TYPE {
+				if !hasComposeNodeTempVar {
```
msujew, line 185: Bug: `hasComposeNodeTempVar` is initialized to `false` and checked here, but never set to `true`. Any grammar interface with two or more non-array composite-typed fields generates `var cn core.CompositeNode` twice in the same function body. The current example grammars have at most one such field per interface, which is why CI is green.

reference.go
```diff
@@ -292,3 +295,147 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+		Err:     errMsg,
+	})
+}
+
+// UnmarshalJSON revives the reference as a JSON object. It inspects properties named 'ref', 'refText', and 'err'.
+// 'ref' is expected to contain a URI string denoting the target node within the same or another document, may be absent if the reference was unresolvable before serializing.
+// 'refText' denotes the cross reference string.
+// 'err' contains the error msg if the reference was unresolvable before serializing
+// With this function Reference[T] implements json.Unmarshaler.UnmarshalJSON() and the method is called by `json.Unmarshal()`.
+func (r *Reference[T]) UnmarshalJSON(data []byte) error {
```
msujew, line 357: Bug: `UnmarshalJSON` has no handling for JSON `null`. A `json.RawMessage` holding an explicit `null` is the *non-nil* bytes `"null"` (encoding/json calls RawMessage's UnmarshalJSON even for null).

Reproduced against examples/statemachine: `{"$type":"Transition","event":null}` creates a phantom reference with empty refText that later fails linking with “Reviving reference in Json document failed”, and an explicit null child-node field fails the whole unmarshal with `unmarshal: unknown type ""`. Roundtrips only pass because MarshalJSON uses `omitempty`. Emitters that write explicit nulls can't be read back. An early `if bytes.Equal(data, []byte("null"))` no-op return here (and in the dispatching Unmarshal) would cover both cases.

util/json/json.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	builder, err := service.Get[workspace.Builder](sc)
+	if err != nil {
+		return err
+	}
+
+	if err := json.Unmarshal(data, rootNode); err != nil {
+		return err
+	}
+	core.AssignContainers(document)
```
msujew, line 32: Bug: `core.AssignContainers(document)` runs before `document.Root = rootNode` is assigned, so it walks the still-nil `Root`.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,295 @@
...
+	node.AppendLine()
+}
+
+func genUnmarshalReference(node codegen.Node, field FieldInfo, srcName string, targetName string, loopItem bool) {
+	genUnmarshalFieldContent(node, field, srcName, targetName, loopItem, func(body codegen.Node) {
+		body.AppendLine(targetName, " := core.NewReference[", strings.Split(field.Type, "[")[1], "(i, nil, nil)")
+		// for the sake simplicity and performance we call 'target.UnmarshalJSON()' directly instead of taking the route
+		// via json.Unmarshal(...), since Reference implements that method
+		// note: the generic impl has special handling for "RawMessage" being equal "null", sets the target pointer to "nil"
+		body.AppendLine("if err := ", targetName, ".UnmarshalJSON(", srcName, "); err != nil {")
```
msujew, line 209: Bug: Generated local identifiers come straight from field names with no collision avoidance against the generator's own temp names; `reservedKeywords` only covers Go keywords. Verified by compiling the generated shapes: a reference field named `Aux` generates `aux := core.NewReference[...]` shadowing the decode struct so the subsequent `aux.Aux` fails with `aux.Aux undefined`; a child field named `Err` produces `err repeated on left side of :=`; and the marshal-side `[]string` temps break for fields named `I`, `J`, or `Item`.

Can we prefix aux variables?

internal/generator/json_generator.go
```diff
@@ -0,0 +1,295 @@
...
+	node.AppendLine(`// Unmarshal decodes data into an instance of type T by reading the "$type" field,`)
+	node.AppendLine(`// selecting a corresponding factory, creating an instance, and unmarshaling its content.`)
+	node.AppendLine("func Unmarshal[T core.AstNode](data []byte) (T, error) {")
+	node.Indent(func(n codegen.Node) {
+		n.AppendLine("node := &struct {")
+		n.Indent(func(n2 codegen.Node) {
+			n2.AppendLine("Type string `json:\"$type\"`")
+		})
+		n.AppendLine("}{}")
+		n.AppendLine("if err := json.Unmarshal(data, node); err != nil {")
```
msujew, line 253: Suggestion: the recursive `Unmarshal[T]` design is O(size \* depth): this `$type` peek runs `json.Unmarshal` over the *entire* subtree at every nesting level just to read one key, each `json.RawMessage` field copies its whole subtree per level, and on the marshal side every node's `json.Marshal` result is re-scanned/compacted by each ancestor.

Since MarshalJSON always emits `$type` first, the peek could be a `json.Decoder` token scan that stops after the first key/value (O(1) instead of O(subtree)), and the unread `T__` fields can simply be dropped. Worth considering given this PR benchmarks exactly these paths.

symbols.go
msujew: This is still unresolved, right?

test/doc_fixture_util.go
```diff
@@ -0,0 +1,138 @@
...
+			getter, exists := expectedValue.Addr().Type().MethodByName(strings.ToUpper(field.Name[0:1]) + field.Name[1:])
+			if !exists {
+				t.Errorf("at %q, type %T: string value getter for field %s missing", p, expectedValue, field.Name)
+			}
+			exp := getter.Func.Call(expectedMethodArg)[0].String()
+			act := getter.Func.Call(actualMethodArg)[0].String()
+			if exp != act {
+				t.Errorf("at %q: primitive string field '%s' mismatch\n  expected: %s\n  actual: %s", p, field.Name, exp, act)
+			}
+		}
```
msujew, line 81: Suggestion: These look identical except for `Bool()` and `String()` calls. Can we simplify this somehow?

test/doc_fixture_util.go
```diff
@@ -0,0 +1,138 @@
...
+		p, _ := core.PathOf(expected)
+		t.Errorf("at %q: Name mismatch: expected %s, got %s", p, expectedNamed.Name(), actualNamed.Name())
+	}
+
+	// 3. Primitive field values must match
+	//  create a reflect.Value of the AstNode:
+	//   * fetch child (1) that is the specific '...Data' struct, child (0) is the 'AstNodeBase' struct
+	//   * then derive a reference value via '.Addr()', and
+	//   * wrap it into an array being used as argument while calling the getters, they're defined with pointer receivers
+	expectedValue := reflect.ValueOf(expected).Elem().FieldByIndex([]int{1})
```
msujew, line 48: Bug: `.FieldByIndex([]int{1})` will only retrieve the first embedded `*Data` struct. However, if a type embeds multiple types (and types like `ParserRule` do that), this comparison only takes the first data struct into account.

internal/generator/json_generator.go
msujew: We have a `NewRootNode()` utility that is supposed to be used here.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,295 @@
...
+				n.AppendLine("i.", field.PName, " = make([]", field.GType, ", 0, len(aux."+field.Name+"))")
+				n.AppendLine("for _, item := range aux.", field.Name, " {")
+				n.Indent(func(n2 codegen.Node) {
+					switch field.GType {
+					case TOKEN_TYPE:
+						n2.AppendLine("i.Set", field.Name, "Item(newToken(Token_ID, item))")
+					case COMPOSITE_TYPE:
+						n2.AppendLine("cn := core.NewCompositeNode()")
+						n2.AppendLine("cn.SetToken(newToken(Token_ID, item))")
+						n2.AppendLine("i.Set", field.Name, "Item(cn)")
```
msujew, line 166: Suggestion: AFAIK with my latest changes, you should be able to use `newToken(nil, item)` here, since a token type is no longer required for a token to be generated. Can we introduce a new constructor in `token.go` for this? `NewSyntheticToken(image string)`?

internal/generator/json_generator.go
msujew: Same here.

util/json/json.go
msujew: I agree.

reference.go
```diff
@@ -292,3 +295,147 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+
+func NewJsonReferenceGetter[T AstNode](uriString string) ReferenceGetter[T] {
+	return func(ctx context.Context, ref *Reference[T]) (*SymbolDescription, *ReferenceError) {
+		if ref.err != nil {
+			return nil, ref.err
+		}
+		if uriString == "" {
+			return nil, NewReferenceError("Reviving reference in Json document failed, 'ref' is absent of empty")
+		}
+		helper, ok := ctx.Value(jsonLinkingHelperKey).(JsonLinkingHelper)
```
msujew, line 412: Suggestion: I already throught about adding the service container to the root context, so it's always available. Maybe that makes more sense (and then retrieve the `DocumentManager` directly here?).

msujew (commented):
Found a few potential bugs looking over this.

internal/generator/json_generator.go
sailingKieler: switched.

internal/generator/json_generator.go
sailingKieler: dropped this function, found a neat way to inline the call of `core.NewToken(...)`.

util/json/json.go
sailingKieler: `util/json/json.go` to `util/json.go`, changed the declared package to `util`.

symbols.go
sailingKieler: Now it is.

internal/generator/json_generator.go
sailingKieler: Kicked out this source of error by wrapping each creation and assignment of a `token` by a pair of `{}`.
This way I can use the same temp variable name each time again.

util/json.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	builder, err := service.Get[workspace.Builder](sc)
+	if err != nil {
+		return err
+	}
+
+	if err := json.Unmarshal(data, rootNode); err != nil {
+		return err
+	}
+	core.AssignContainers(document)
```
sailingKieler: Indeed. 🤔
Fixed.

reference.go
sailingKieler: That's also an option.
However, I assume there's a notable overhead of fetching the `DocumentManager` for each and every cross-ref , isn't it?

internal/generator/json_generator.go
sailingKieler: I reworked the code generation wrt. this, added a minimal test language to identify potential issues quickly.

internal/generator/json_generator.go
sailingKieler: Done (although I'm not really convinced), but also left a sugar alias here.

internal/generator/json_generator.go
sailingKieler: Implement this optimization, reduces unmarshaling time by roughly 40%, which makes sense as for each included astNode one of two full lexing runs is omitted.

internal/generator/json_generator.go
sailingKieler: Added `NewSyntheticToken(image string)` as it makes sense from my POV, too.

test/doc_fixture_util.go
sailingKieler: Tried it. It's different now, not really simpler..

test/doc_fixture_util.go
sailingKieler: Very good catch!
I reworked this part, works now. The impl now creates sub tests for each field, which may slow down the execution a bit but gives us a nice insight which fields are considered.

reference.go
sailingKieler: `JsonLinkingHelper` is currently required, added the missing documentation.
I removed superfluous definitions, `NewJsonReferenceGetter` is private now.

Public defs are documented now.

util/json.go
```diff
@@ -0,0 +1,61 @@
...
+
+	core "typefox.dev/fastbelt"
+	"typefox.dev/fastbelt/util/service"
+	"typefox.dev/fastbelt/workspace"
+)
+
+// UnmarshalAndBuildDocument uses the "encoding/json" entry point to unmarshal rootNode based on the given data string and builds the document.
+// Similar to parsing-based document loading the ast build up first including the reference, while the resolution of the references is done during the linking phase of the building process.
+// For properly linking references to other documents, a helper object is attached to the context given to builder providing access to other documents.
+func UnmarshalAndBuildDocument[T core.AstNode](sc *service.Container, document *core.Document, rootNode T, data []byte, ctx context.Context) error {
```
sailingKieler: > The ctx parameter should be the first per context package convention.

It is now.

reference.go
sailingKieler: Also a good point.
Added such a check in the caller of this method, as well as for AstNodes.

reference.go
sailingKieler: That's the only point I did not applied, yet.
Let's talk about a good solution.

sailingKieler:
This PR is now in a shape that I like a lot. 🥳
I cared about all remarks and suggestions, except for https://github.com/TypeFox/fastbelt/pull/98#discussion_r3842910395 that I would like to talk about first.

I the course of switching to https://pkg.go.dev/encoding/json/v2 I had to increase the Go version to 1.27.
This causes some follow-up changes.
In order to keep proper debugging in VS Code devs need to update their version of [dlv](https://github.com/go-delve/delve/) to [v1.27.1](https://pkg.go.dev/github.com/go-delve/delve/cmd/dlv@v1.27.1) via
```
go install github.com/go-delve/delve/cmd/dlv@v1.27.1 // or @latest
```

@msujew looking forward to your feedback.

util/json.go
```diff
@@ -0,0 +1,179 @@
...
+
+// UnmarshalValue reads the given JSON value into an instance of type T by
+// reading the "$type" field, selecting a corresponding factory, creating an instance, and
+// unmarshaling its content via the decoder API of package "encoding/json/v2".
+//
+// Returns an error if the "$type" field is missing or unknown, or if the instance cannot be converted to type T.
+func UnmarshalValue[T core.AstNode](value jsontext.Value, factories map[string]func() core.AstNode) (T, error) {
+	var zero T
+	// first check for an explicit "null" value
+	if bytes.Equal(value[0:4], []byte("null")) {
```
msujew: `value[0:4]` panics for any JSON value shorter than four bytes. `arithmetics import` on a file containing `{}`, `[]`, `1` or `""` gives `slice bounds out of range [:4] with capacity 2`, because `ReadValue` returns a cap-limited slice. For nested short values (`"statements":[{}]`) the copy has cap >= 8, so this silently reads zeroed bytes past `len` and falls through to a confusing `unknown type ""` error.

`value.Kind() == 'n'` is the safe stdlib check. Same pattern at line 117 in `UnmarshalReference`.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,279 @@
...
+	field FieldInfo,
+	srcName string,
+	targetName string,
+	thisDotSet string,
+	loopItem bool,
+	unmarshalBody func(codegen.Node),
+) {
+	if loopItem {
+		unmarshalBody(node)
+		node.AppendLine(thisDotSet, field.Name, "Item(", targetName, ")")
```
msujew: The generated list loop appends whatever `UnmarshalValue` / `UnmarshalReference` returns, including the nil they return for an explicit `null` item. Both paths crash:

- `{"$type":"Module","name":"m","statements":[null]}` -> `SetStatementsItem(nil)` -> `AssignContainers` panics at `ast.go:391` (`child.SetDocument` on nil).
- statemachine `"actions":[null]` -> typed-nil `*Reference` in `doc.References` -> `CreateLinkerDiagnostics` calls `Reference.TextRange` on a nil receiver inside a builder worker goroutine, taking the process down.

Skipping nil items in the generated loops fixes both.

reference.go
```diff
@@ -289,3 +293,153 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+
+func newJsonReferenceGetter[T AstNode](uriString string) ReferenceGetter[T] {
+	return func(ctx context.Context, ref *Reference[T]) (*SymbolDescription, *ReferenceError) {
+		if ref.err != nil {
+			return nil, ref.err
+		}
+		if uriString == "" {
+			return nil, NewReferenceError("Reviving reference in Json document failed, 'ref' is absent of empty")
+		}
+		helper, ok := ctx.Value(jsonLinkingHelperKey).(JsonLinkingHelper)
```
msujew, line 431: The helper lookup runs before the same-document `#/...` branch at line 424, so every revived reference needs `JsonLinkingHelper` in the build context, even intra-document ones. `DefaultDocumentUpdater` builds under a context derived from `context.Background()`, so the first edit in a workspace containing a JSON-loaded document relinks it without the helper.

Reproduced: load `price-calc.calc.json` via `UnmarshalAndBuildDocument` (13 refs link fine), then `builder.Reset` + `builder.Build(context.Background(), ...)` -> 13/13 fail with `JsonLinkingHelper unavailable`.

Related: `Reference.Reset` nils `r.err`, so a reference imported with `$error` and rebuilt (even with the helper present) changes its message to `'ref' is absent of empty`. Storing the document lookup and the serialized error in the getter closure instead of ctx / `r.err` removes both problems.

util/json.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return err
+	}
+
+	if document.Root, err = Unmarshal[core.AstNode](input, factories); err != nil {
+		return err
+	}
+
+	document.State = core.DocStateParsed
+	core.AssignContainers(document)
```
msujew, line 149: `Unmarshal` returns `(nil, nil)` for a top-level `null`, so `document.Root` is nil here and `AssignContainers` panics with a nil pointer dereference (`echo null | arithmetics import` reproduces it). Worth returning an error when the root is nil.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,279 @@
...
+				})
+				n.AppendLine("}")
+			} else if field.Boolean {
+				n.AppendLine("if aux.", field.Name, "{")
+				n.Indent(func(n2 codegen.Node) {
+					n2.AppendLine(thisDotSet, field.Name, "(", genCreateNewToken("\"\""), ")")
+				})
+				n.AppendLine("}")
+			} else if field.GType == TOKEN_TYPE {
+				n.AppendLine(thisDotSet, field.Name, "(", genCreateNewToken("aux."+field.Name), ")")
```
msujew: The aux field type for token/composite fields is plain `string`, so an omitted optional property is indistinguishable from `""` and this always creates a token. After parsing the same token is nil.

Reproduced on the grammar language: `token ID:` has `TypeToken() == nil` after parsing, but after a marshal/unmarshal round trip it is non-nil with image `""` and range `{-1,-1}`. `linking.Name` then returns a non-nil unit, so `DescribeExport` / document symbols export an empty-named symbol at a -1..-1 range, and `core.WithToken` sets diagnostic ranges to `{-1,-1}`.

`AssertEqualAst` only compares getter strings, so `TestJsonRoundtrip` cannot see this. A pointer / `omitzero` aux field with a nil check would keep absence intact.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,279 @@
...
+		// instead of implementing 'MarshalJSON()' for *core.Token, a preprocessing loop calling '.String()' for each item
+		// is added for each of such fields
+		var stringListFields = map[string]string{}
+		for _, field := range fields {
+			if field.Array && (field.GType == TOKEN_TYPE || field.GType == COMPOSITE_TYPE) {
+				varName := field.PName
+				stringListFields[field.Name] = varName
+
+				n.AppendLine(varName, " := make([]string, len(", thisDot, field.Name, "()))")
+				n.AppendLine("for j, item := range ", thisDot, field.Name, "() {")
```
msujew: `varName := field.PName` is grammar-derived, but the loop below uses the fixed locals `item` and `j`. Generating from `interface Coll { Item []string  J []string }` emits

```go
for j, item := range _this.Item() { item[j] = item.String() }
for j, item := range _this.J() { j[j] = item.String() }
```

neither of which compiles. The collision probe grammar in `internal/languages/json/json.fb` covers `aux/core/err/node/this/util/type/token` but not `item` or `j`. Using reserved-prefix names (like `_this` / `_err`) for every generator-owned local closes the whole class.

reference.go
```diff
@@ -289,3 +293,153 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+		if !ok {
+			return nil, &ReferenceError{
```
msujew: This is the only `ReferenceError` built as a literal; `NewReferenceError` sets `SeverityError`, the literal leaves `Severity` at 0 ("Unknown", outside the LSP 1-4 range). `CreateLinkerDiagnostics` copies it straight through, and both CLIs count only `Severity == SeverityError`, so a document whose references all failed this way is printed but exits 0.

symbols.go
```diff
@@ -125,6 +125,12 @@ func (c *mergedSymbolContainer) All() SymbolSeq {
 
 func (c *mergedSymbolContainer) ForType(targetType reflect.Type) SymbolSeq {
 	return extiter.FlatMap(c.containers, func(container SymbolContainer) SymbolSeq {
+		if container == nil {
```
msujew: The nil guard was added here in `ForType` but `All()` (line 122) still calls `container.All()` on every element. nil really gets in: `ImportSymbols` maps every document's `ExportedSymbols`, which is nil until built and after `Reset`.

Reproduced: `Set` an unbuilt document into the manager, build another one, then `doc.ImportedSymbols.All()` panics while `ForType` works. No in-repo caller uses `All()` on the merged container today, so it is latent for client scope/completion providers. Filtering nil once at the `MergeSymbolContainers` site would cover both methods.

test/fixture.go
```diff
@@ -183,12 +183,22 @@ func (f *Fixture) ParseAll(uriContentPairs ...string) []*Doc {
...
 	f.docs = append(f.docs, testDoc)
 	return testDoc
 }
 
+func (f *Fixture) Delete(uri core.URI) {
+	documents := service.MustGet[workspace.DocumentManager](f.sc)
+	documents.Delete(uri)
+}
+
+func (f *Fixture) Clear() {
```
msujew, line 202: `Clear` and `Delete` remove documents from the `DocumentManager` but never prune `f.docs`, so `Fixture.Documents()` and the LSP helpers that iterate `d.fixture.docs` still see deleted documents.

This already affects `examples/arithmetics/json_test.go`: the loops at lines 129 and 167 iterate `f.Documents()` and see 5 and 7 documents while the manager holds 2. The stale documents keep their old linked state and pass vacuously, so the assertions are silently weaker than they look.

`Delete` also has no callers in the repo.

server/diagnostics_publisher.go
```diff
@@ -106,7 +107,7 @@ func toLspDiagnostic(textDoc textdoc.Handle, d core.Diagnostic) lsp.Diagnostic {
 	if d.Data != nil {
 		raw, err := json.Marshal(d.Data)
```
msujew: Was the switch to json v2 here intended? `lsp.Diagnostic.Data` is `*json.RawMessage`, which is `= jsontext.Value` on go1.27, so the v1 import compiled unchanged. v2 changes the wire encoding of user-supplied `Data`: nil slices become `[]`, `omitempty` no longer drops `false`/`0`, map keys are unsorted, and invalid UTF-8 returns an error instead of being replaced. The `if err == nil` guard then silently drops `Data` entirely.

If v1 behaviour should be preserved, `json.Marshal(d.Data, jsonv1.DefaultOptionsV1())` reproduces the v1 output byte-for-byte. Only downstream `WithData` users are affected; there are no in-repo callers.

examples/arithmetics/cli/main.go
```diff
@@ -0,0 +1,208 @@
...
+		Short:   "Import an arithmetics description from JSON",
+		Long:    importLongHelp,
+		Example: importExamples,
+		Args:    cobra.ExactArgs(1),
+		RunE: func(_ *cobra.Command, args []string) error {
+			opts.inputPath = args[0]
+			return runImportCLI(opts)
+		},
+	}
+	cmd.Flags().StringVarP(&opts.outputPath, "output", "o", "", "Output file path (defaults to stdout)")
```
msujew, line 110: `import` registers `-o/--output` and the long help says "Use --output to write to a file instead", but `runImportCLI` never reads `opts.outputPath` and always writes to stdout. `arithmetics import x.json -o out.arithmetics` exits 0, prints to stdout and creates no file.

`importOptions` / `newImportCmd` are otherwise verbatim copies of the export pair, so one options struct and one command builder could serve both, with the flag either honored or dropped.

util/json.go
```diff
@@ -0,0 +1,179 @@
...
+	if !ok {
+		return zero, fmt.Errorf("util.UnmarshalValue: unknown type %q", nodeType)
+	}
+	instance := factory()
+	asT, ok := instance.(T)
+	if !ok {
+		return zero, fmt.Errorf("util.UnmarshalValue: %T is not convertible to type %s", instance, reflect.TypeFor[T]())
+	}
+	if unmarshaler, ok := instance.(json.UnmarshalerFrom); ok {
+		if err := unmarshaler.UnmarshalJSONFrom(jsontext.NewDecoder(bytes.NewReader(value))); err != nil {
```
msujew: A fresh `jsontext.Decoder` over a `bytes.Reader` is built for every AST node (and again in `peekTypeFast`), which allocates and copies the subtree per node. json v2 already dispatches to `UnmarshalerFrom` for pointer types and uses a pooled decoder that aliases the input.

Measured on `BenchmarkJsonRoundtrip/Fastbelt_Grammar_language/unmarshal-self`:

| variant | ns/op | B/op | allocs/op |
|---|---|---|---|
| HEAD | 2.6 ms | 2.36 MB | 16550 |
| `json.Unmarshal(value, instance)` | 0.93 ms | 528 KB | 7395 |
| plus pooled peek decoder | 0.67 ms | 320 KB | 4372 |

All tests pass with the change. It requires the field-less `UnmarshalJSONFrom` fix first (see the comment at `json_generator.go:132`), since `json.Unmarshal` enforces the read-exactly-one-value contract.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,279 @@
...
+
+	node.AppendLine("func (_this *", iface.Name(), "Impl) UnmarshalJSONFrom(_decoder *jsontext.Decoder) error {")
+	thisRef := "_this"
+	thisDotSet := thisRef + ".Set"
+	errRef := "_err"
+	genCreateNewToken := func(valueRef string) string {
+		return "core.NewSyntheticToken(" + valueRef + ", " + thisRef + ")"
+	}
+
+	if len(fields) == 0 {
```
msujew, line 172: For interfaces with zero fields the generated `UnmarshalJSONFrom` is `return nil` without reading the value, which violates json v2's `UnmarshalerFrom` contract. It only works today because `UnmarshalValue` hands each node a throwaway decoder.

``json.Unmarshal([]byte(`{"$type":"Expression"}`), arithmetics.NewExpression())`` fails with `must read or write exactly one value`; same for `StatementImpl` and any user grammar marker node. Emitting `_, err := _decoder.SkipValue(); return err` fixes it.

examples/arithmetics/cli/main.go
```diff
@@ -0,0 +1,208 @@
...
+		return err
+	}
+
+	inputText, err := os.ReadFile(inputPath)
+	if err != nil {
+		return err
+	}
+
+	sc := arithmetics.CreateServices()
+	file, _ := textdoc.NewFile(lsp.URIFromPath(inputPath), "arithmetics", 0, string(inputText))
```
msujew, line 130: Lines 123-161 are a near-verbatim copy of `cmd/fastbelt/generate.go:57-97`, and `textdoc.NewFile` + `core.NewDocument` re-implement `core.NewDocumentFromString`. The copy has already drifted (see the diagnostics format below). A shared load-and-report helper next to `util.UnmarshalAndBuildDocument` would serve both mains.

Also, `runImportCLI` at line 190 hand-builds the URI as `"file:///" + filepath.Base(inputPath)`, dropping the directory, while export uses `lsp.URIFromPath(inputPath)` here.

reference.go
```diff
@@ -289,3 +293,153 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+		refDoc := referenced.Document()
+		if refDoc == nil {
+			referencerPath, _ := PathOf(r.Owner())
+			return fmt.Errorf("Reference.MarshalJSONTo(): node of type %T referenced by '%s' is not contained in any document", referenced, referencerPath)
+		}
+
+		if r.Owner() != nil && r.Owner().Document() == refDoc {
+			uri = "#" + refPath.String()
+		} else {
+			uri = refDoc.URI.WithFragment(refPath.String()).StringUnencoded()
```
msujew: Cross-document targets are written with `StringUnencoded()` but revived with `ParseURI`, which splits on `#` and percent-decodes. The `DocumentManager` keys by `StringUnencoded()`, so paths containing `%` or `#` do not round-trip:

- `C:\proj\a%20b\x.calc` exports `file:///C:/proj/a%20b/x.calc#/statements@0`; `ParseURI` decodes it to `.../a b/x.calc`, the lookup misses and every cross-doc reference reports `Document not found`.
- `C:\proj\#1\x.calc` splits at the first `#`, corrupting both the path and the node path.

Emitting `String()` (encoded) round-trips all cases; the same-document branch at line 331 and the error strings around line 429 would want the same treatment.

reference.go
```diff
@@ -289,3 +293,153 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+
+		if document.Root == nil {
+			return nil, NewReferenceError("Document is empty: '" + document.URI.StringUnencoded() + "'")
+		}
+
+		node, err := Resolve(document.Root, uri.Fragment())
+		if err != nil {
+			return nil, NewReferenceError(err.Error())
+		}
+		return &SymbolDescription{
```
msujew: This leaves `Name` nil and `URI` as the raw `$ref` (`#/statements@0` for same-document refs), unlike `NewSymbolDescription`, which sets the target document URI and the name node. After a JSON build every reference's `Description()` looks like this (207/207 on the grammar language).

`server/name_finder.go:73-77` yields `Target == nil` for such references, so hover/definition/rename would report not found on JSON-imported documents once they carry text. Latent today because imported documents have no text. `core.NewSymbolDescription(node, name)` gives the same shape as the rest of the linker.

util/json.go
```diff
@@ -0,0 +1,179 @@
...
+		),
+		[]*core.Document{document}, nil,
+	); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+type defaultJsonLinkingHelper struct {
```
msujew: `defaultJsonLinkingHelper.GetDocument` is a one-line forward to `workspace.DocumentManager.Get` with an identical signature. If `core.JsonLinkingHelper`'s method were named `Get`, any `DocumentManager` would satisfy it structurally and lines 162-179 plus the wrapping at the three call sites (`util/json.go:152`, both json tests) could go.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,279 @@
...
+	thisRef string,
+	thisDotSet string,
+	errRef string,
+	loopItem bool,
+) {
+	genUnmarshalFieldContent(node, field, srcName, targetName, thisDotSet, loopItem, func(body codegen.Node) {
+		// for the sake simplicity and performance we call 'target.UnmarshalJSON()' directly instead of taking the route
+		// via json.Unmarshal(...), since Reference implements that method
+		// note: the generic impl has special handling for "RawMessage" being equal "null", sets the target pointer to "nil"
+		body.AppendLine(targetName, ", ", errRef, " := util.UnmarshalReference[", strings.Split(field.Type, "[")[1], "(", thisRef, ", ", srcName, ")")
```
msujew: `strings.Split(field.Type, "[")[1]` yields `"X]"` from `*core.Reference[X]` and relies on the trailing `]` closing the emitted `util.UnmarshalReference[X](` call. The target name is already in hand as `refType.Type().Text()` inside `getTypeName`; storing it on `FieldInfo` avoids re-deriving it from the rendered Go type, which would silently break if the rendering ever changes.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,279 @@
...
+	loopItem bool,
+	unmarshalBody func(codegen.Node),
+) {
+	if loopItem {
+		unmarshalBody(node)
+		node.AppendLine(thisDotSet, field.Name, "Item(", targetName, ")")
+	} else {
+		node.AppendLine("if ", srcName, " != nil {")
+		node.Indent(func(n2 codegen.Node) {
+			unmarshalBody(node)
```
msujew: Two things here:

- The `Indent` callback ignores `n2` and writes to the outer `node`, so the body is emitted at the wrong indent. Pre-gofmt output shows `if aux.Other != nil {` followed by body lines at the same column; `FormatIfPossible` repairs it unless formatting fails.
- `genUnmarshalReference` and `genUnmarshalChild` differ only in one body line and thread the constants `"_this"`, `"_this.Set"`, `"_err"` through 7-8 positional params. A single `genUnmarshalField(node, field, src, target, loopItem, callExpr)` with package-level constants would remove the duplication.

test/doc_fixture_util.go
```diff
@@ -0,0 +1,161 @@
...
+	expectedGetterArg := []reflect.Value{expectedValueRef}
+
+	actualValueRef := reflect.ValueOf(actual)
+	actualGetterArg := []reflect.Value{actualValueRef}
+
+	for subStructIndex := 1; subStructIndex != expectedValue.NumField(); subStructIndex++ {
+		fields := expectedValue.FieldByIndex([]int{subStructIndex}).Type().Fields()
+
+		for field := range fields {
+			isToken := field.Type.Kind() == reflect.Pointer && field.Type == REF_TOKEN_TYPE
```
msujew, line 71: `AssertEqualAst` compares only fields of exact type `*core.Token`. `[]*core.Token`, `CompositeNode` and `*core.Reference` fields are never compared, and `TestJsonRoundtrip` does not assert that the re-imported document has no linking errors. An off-by-one `$ref` path or a dropped `regexps` list would leave the round-trip tests green.

Comparing reference targets (or asserting no linking errors on the imported doc) plus list/composite images would close that gap. Minor: the `Kind() == reflect.Pointer` check is implied by the type equality that follows.

test/doc_fixture_util.go
```diff
@@ -0,0 +1,161 @@
...
+					stringGetterName = stringGetterName[1:]
+				}
+
+				getter, exists := expectedValueRefType.MethodByName(stringGetterName)
+				if !exists {
+					// if no 'string' getter exists, look for a 'bool' getter,
+					//  we can't distinguish this on the struct field decls
+					getter, exists = expectedValueRefType.MethodByName("Is" + stringGetterName)
+					if !exists {
+						t.Errorf("at %q, type %T: value getter for field %s missing", pExpected, expected, field.Name)
```
msujew, line 90: When neither getter exists this reports via `t.Errorf` but does not return, so line 93 calls `getter.Func.Call` on a zero `reflect.Method` and panics with `reflect: call of reflect.Value.Call on zero Value`, aborting the whole comparison. Unreachable with current generator output (all 64 token fields across the 7 languages have getters), so just a missing `return`.

test/doc_fixture_util.go
```diff
@@ -0,0 +1,161 @@
...
+
+		for field := range fields {
+			isToken := field.Type.Kind() == reflect.Pointer && field.Type == REF_TOKEN_TYPE
+			if !(isToken) {
+				continue
+			}
+
+			// for each field create a sub test, use the composed type's name instead of the component type's name
+			// to allow us seeing immediately that the all fields of the composed types are hit
+			t.Run(expectedValueType.Name()+"."+field.Name, func(t *testing.T) {
```
msujew, line 78: A `t.Run` per token field of every node makes the harness dominate: `go test ./internal/grammar/ -run TestJsonRoundtrip -v` emits 2402 lines for 1200 subtests over four grammars, while the comparison itself is a few reflect calls per node. The two `iter.Pull` coroutines per node (line 130) and the eager `core.PathOf` at line 26 (only used in error branches) add to it.

Plain `t.Errorf` with the field name in the message, an index loop over the two child slices, and a lazily computed path would drop the overhead and the `iter`/`slices` imports.

internal/grammar/json_test.go
```diff
@@ -0,0 +1,140 @@
...
+			require.NoError(t, err)
+
+			exported, imported := parseExportImportGrammar(t, grammmar)
+			test.AssertEqualAst(t, exported, imported) // expected := exported, actual := imported
+		})
+	}
+}
+
+func BenchmarkJsonRoundtrip(b *testing.B) {
+	tests := []struct {
```
msujew: This table is byte-identical to the one in `TestJsonRoundtrip` at line 23; a package-level var would serve both. The `b.ResetTimer()` calls at lines 69, 77 and 86 immediately precede `b.Loop()`, which resets the timer itself, so they are dead.

Related: `loadJsonDoc` in `examples/arithmetics/json_test.go:96` repeats the same Set-without-build document wiring as lines 127-132 here. A shared fixture helper would fit both (not `UnmarshalAndBuildDocument`, which builds immediately).

examples/arithmetics/cli/main.go
```diff
@@ -0,0 +1,208 @@
...
+	})
+
+	for _, diag := range diagnostics {
+		if diag.Severity == core.SeverityError {
+			errCount++
+		}
+		tRange := diag.Range.LspRange(document.TextDoc)
+		fmt.Fprintf(os.Stderr, "%s - %d:%d %s\n",
+			diag.Severity.String(),
+			tRange.Start, tRange.End,
```
msujew: `tRange.Start` and `tRange.End` each format as `line:char`, so this prints four 0-based numbers where the format suggests one `line:col` pair. An error at line 2 column 11 prints `Error - 1:10:1:11 ...`, whereas `generate.go` prints 1-based `Start.Line+1, Start.Character+1`. The error is also printed twice because cobra's `SilenceErrors` is unset while `main` prints it too.

examples/arithmetics/cli/main.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	return os.WriteFile(opts.outputPath, append(out, '\n'), 0644)
+}
+
+func runImportCLI(opts importOptions) error {
+	inputPath, err := filepath.Abs(opts.inputPath)
+	if err != nil {
+		return err
+	}
+	file, err := os.Open(inputPath)
```
msujew, line 193: The file is never closed (neither here nor in `UnmarshalAndBuildDocument`, which takes an `io.Reader`). Harmless for a one-shot CLI, but `defer file.Close()` after the error check, or `os.ReadFile` + `bytes.NewReader`.

reference.go
```diff
@@ -289,3 +293,153 @@ func NewReferenceDescriptionsFromMap(descriptions collections.MultiMap[AstNode,
...
+		Ref     string `json:"$ref,omitempty"`
+		Err     string `json:"$error,omitempty"`
+	}{
+		RefText: r.unit.String(),
+		Ref:     uri,
+		Err:     errMsg,
+	})
+}
+
+// UnmarshalJSON revives the reference as a JSON object. It inspects properties named 'ref', 'refText', and 'err'.
```
msujew: The comment describes properties `'ref'`, `'refText'` and `'err'`, but the struct tags are `"$ref"`, `"$refText"` and `"$error"`. It also says `json.Unmarshal` calls this method, while `util.UnmarshalReference` calls it directly. Someone hand-writing JSON from the comment gets an unresolvable reference.

The user-visible error at line 414 reads `'ref' is absent of empty` (typo for "or"); it reaches re-marshaled `$error` output.

internal/generator/json_generator.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+		n.AppendLine("return json.MarshalEncode(_encoder, struct {")
+		n.Indent(func(n2 codegen.Node) {
+			n2.AppendLine("T__", " ", "string", " `json:\"$type\"`")
+			for _, field := range fields {
+				typeStr := field.Type
+				if field.Array {
+					typeStr = "[]" + typeStr
+				}
+				n2.AppendLine(field.Name, " ", typeStr, " `json:\"", field.JsonPropName, ",omitempty\"`")
```
msujew: Under json v2 `omitempty` only drops values that encode as `null`, `""`, `{}` or `[]`; `false` is always emitted. So every optional `?=` assignment always appears as `"x": false` in exported JSON, unlike v1 and unlike Langium's output. If omission is intended for bools, `omitzero` is the v2 spelling.

go.mod
```diff
@@ -1,6 +1,6 @@
 module typefox.dev/fastbelt
 
-go 1.26.0
+go 1.27.0
```
msujew, line 3: Suggestion: Please also update the devcontainer.json version.

msujew (commented):
Claude found a few additional issues. I've removed some (since they weren't really relevant) - but others can actually crash/break the serialization.

internal/generator/json_generator.go
sailingKieler: I didn't expected `null` values in lists at all, and I agree to add this for the sake of robustness.

internal/generator/json_generator.go
sailingKieler: Agree, although it's just cosmetics.

reference.go
sailingKieler: Added clarifying comments.
The question on the `JsonLinkingHelper` is not decided finally, so the current state is preliminary in that only regard.

internal/generator/json_generator.go
sailingKieler: Solved, including an additional test within the `Arithmetics` language impl.

symbols.go
sailingKieler: Changed.

examples/arithmetics/cli/main.go
sailingKieler: I don't care. Claude changed something...

internal/generator/json_generator.go
sailingKieler: Fixed.

internal/generator/json_generator.go
sailingKieler: Fixed.

util/json.go
sailingKieler: Impressive!

reference.go
sailingKieler: fixed.

reference.go
sailingKieler: done.

test/doc_fixture_util.go
sailingKieler: done.

internal/grammar/json_test.go
sailingKieler: consolidated.

symbols.go
```diff
@@ -90,21 +90,27 @@ func (c *emptySymbolContainer) Put(desc *SymbolDescription) bool {
...
+	return EmptySymbolSeq
 }
 
 // MergeSymbolContainers merges multiple symbol containers into one. The resulting container
 // is immutable and reflects the combined contents of all input containers.
 func MergeSymbolContainers(containers iter.Seq[SymbolContainer]) SymbolContainer {
-	if extiter.IsEmpty(containers) {
+	// A container may be 'nil', e.g. in case a document in the document manager hasn't been built
+	// yet or has been reset and is not included in the current build cycle, so its exported symbols
+	// container is absent; filter such entries out once here rather than in every method below.
```
msujew: Suggestion: This comment is inaccurate. Filtering happens on every `ForType`, not just once here (`Filter` is lazy, not eager). Either adjust the comment, or make the filtering eager here (by collecting it into an array).

msujew (approved):
I think this is good to go now. Please have a look at the comment below before merging.

PR #99, opened 2026-06-08
First hover and documentation providers implementation

gfontorbe:
Closes #52 

First attempt at implementing a default hover and documentation provider.

The Hover provider responds to a Hover Request by calling the Documentation Provider to get the content of the associated comment if it exists.

The Documentation Provider finds the comment and strips the comment tokens and returns a cleaned-up string.

I did not find how to get the exact tokens used for comments so I defaulted to removing and leading and trailing non-alphanumerical character. Open for suggestions on how to improve.

### DIFF
--- a/examples/statemachine/hover_test.go
+++ b/examples/statemachine/hover_test.go
@@ -1,0 +1,224 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+package statemachine
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"typefox.dev/fastbelt/server"
+	"typefox.dev/fastbelt/test"
+	"typefox.dev/fastbelt/util/service"
+	"typefox.dev/lsp"
+)
+
+// createHoverServices builds a service container with both the statemachine language
+// and the default server services (DocumentationProvider, HoverProvider, etc.).
+func createHoverServices() *service.Container {
+	sc := service.NewContainer()
+	SetupServices(sc)
+	server.SetupDefaultServices(sc)
+	sc.Seal()
+	return sc
+}
+
+// hoverAt calls the hover provider at the position of the named marker.
+// Accepts both index markers (<|label>) and range markers (<|label:text|>);
+// for range markers the start of the range is used as the cursor position.
+func hoverAt(t *testing.T, f *test.Fixture, doc *test.Doc, label string) *lsp.Hover {
+	t.Helper()
+	offset := -1
+	for _, idx := range doc.Indices {
+		if idx.Label == label {
+			offset = idx.Offset
+			break
+		}
+	}
+	if offset == -1 {
+		for _, r := range doc.Ranges {
+			if r.Label == label {
+				offset = r.Start
+				break
+			}
+		}
+	}
+	if offset == -1 {
+		t.Fatalf("no marker with label %q", label)
+	}
+	pos := doc.Document.TextDoc.PositionAt(offset)
+	provider := service.MustGet[server.HoverProvider](f.Services())
+	result, err := provider.HandleHoverRequest(f.Ctx(), &lsp.HoverParams{
+		TextDocumentPositionParams: lsp.TextDocumentPositionParams{
+			TextDocument: lsp.TextDocumentIdentifier{URI: doc.Document.URI.DocumentURI()},
+			Position:     pos,
+		},
+	})
+	require.NoError(t, err)
+	return result
+}
+
+func TestDocumentationSingleLineComment(t *testing.T) {
+	f := test.New(t, createHoverServices())
+	docProvider := service.MustGet[server.DocumentationProvider](f.Services())
+
+	doc := f.Parse(`
+statemachine Test
+events flick
+initialState off
+
+// The off state
+state off
+  flick => on
+end
+
+state on
+  flick => off
+end
+`).AssertNoErrors()
+
+	off := test.MustFindNamedNode[State](doc, "off")
+	assert.Equal(t, "The off state", docProvider.Documentation(off))
+}
+
+func TestDocumentationMultipleLineComments(t *testing.T) {
+	f := test.New(t, createHoverServices())
+	docProvider := service.MustGet[server.DocumentationProvider](f.Services())
+
+	doc := f.Parse(`
+statemachine Test
+events flick
+initialState off
+
+// line one
+// line two
+state off
+  flick => on
+end
+
+state on
+  flick => off
+end
+`).AssertNoErrors()
+
+	off := test.MustFindNamedNode[State](doc, "off")
+	assert.Equal(t, "line one  \nline two", docProvider.Documentation(off))
+}
+
+func TestDocumentationBlockComment(t *testing.T) {
+	f := test.New(t, createHoverServices())
+	docProvider := service.MustGet[server.DocumentationProvider](f.Services())
+
+	doc := f.Parse(`
+statemachine Test
+events flick
+initialState off
+
+/*
+ * line one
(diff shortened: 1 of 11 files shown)


examples/statemachine/hover_test.go
```diff
@@ -0,0 +1,224 @@
...
+	SetupServices(sc)
+	server.SetupDefaultServices(sc)
+	sc.Seal()
+	return sc
+}
+
+// hoverAt calls the hover provider at the position of the named marker.
+// Accepts both index markers (<|label>) and range markers (<|label:text|>);
+// for range markers the start of the range is used as the cursor position.
+func hoverAt(t *testing.T, f *test.Fixture, doc *test.Doc, label string) *lsp.Hover {
```
msujew: Suggestion: Can we move this functionality into the `doc_fixture_lsp.go` file? Ideally just like `ExpectHoverAt(label, markup string)`.

server/documentation_provider.go
```diff
@@ -0,0 +1,100 @@
...
+		commentEnd := int(c.TextSegment.Indices.End)
+		if commentEnd > prevEnd {
+			continue
+		}
+		if hasTokenInRange(doc.Tokens, commentEnd, prevEnd) {
+			break
+		}
+		block = append(block, c)
+		prevEnd = int(c.TextSegment.Indices.Start)
+	}
```
msujew: Suggestion: This feels like it could become quite expensive on large documents with lots of consecutive comments. Maybe we can do something like this:
1. Find the index of the first token associated with `node` (note: return `""` if `len(node.Tokens) == 0`)
2. Use this together with the token at `index-1` to delimit the comment search range.
3. Collect all comment tokens in between the two tokens into `block`.

This way, we only need two binary searches at most (while this needs `n` binary searches with the amount of `n` determined by the amount of comments before `node`).

server/documentation_provider.go
```diff
@@ -0,0 +1,100 @@
...
+	// Reverse to restore chronological order.
+	for i, j := 0, len(block)-1; i < j; i, j = i+1, j-1 {
+		block[i], block[j] = block[j], block[i]
+	}
+
+	var lines []string
+	for _, c := range block {
+		if line := stripCommentMarkers(c.Image); line != "" {
+			lines = append(lines, line)
+		}
```
msujew: Suggestion: I'm not sure the `line != ""` filter is a good idea. For example, if we want to explicitly create a new paragraph like:

```go
// Some description of func
//
// Note: Do not use
func ...
```

The second line gets completely ignored, which results in just a single paragraph in the output.

server/documentation_provider.go
```diff
@@ -0,0 +1,100 @@
...
+
+// stripCommentMarkers removes comment markers from a raw comment token image.
+// Leading and trailing non-alphanumeric runes are trimmed from the whole image
+// and then again from each individual line, stripping interior markers (e.g. the
+// * in /* */ blocks). Works for any comment style (// # -- /* */ etc.).
+func stripCommentMarkers(raw string) string {
+	isMarker := func(r rune) bool {
+		return !unicode.IsLetter(r) && !unicode.IsDigit(r)
+	}
+	raw = strings.TrimFunc(strings.TrimSpace(raw), isMarker)
```
msujew: Good idea in theory, but falls apart in practice. This will also strip all periods `.` at the end of a line.

My suggestion would be to approach this similar to the folding range provider: Add an interface like this that can be optionally supplied to the `DocumentationProvider`:

```go
type DocumentationTrimmer interface {
  TrimComment(content string) string
}
```

The default implement should remove `//` from single line comments and `/*, *, /*` from multi-line comments.

server/hover_provider.go
```diff
@@ -0,0 +1,65 @@
...
+
+	offset := doc.TextDoc.OffsetAt(params.Position)
+	sourceToken := doc.Tokens.SearchOffset(offset)
+	if sourceToken == nil {
+		return nil, nil
+	}
+
+	nameFinder := service.MustGet[NameFinder](s.sc)
+	foundName := nameFinder.Find(ctx, sourceToken)
+	if foundName.Target == nil {
```
msujew: Suggestion: We need both later on.

```suggestion
	if foundName.Target == nil || foundName.Source == nil {
```

server/hover_provider.go
```diff
@@ -0,0 +1,65 @@
...
+		return nil, nil
+	}
+
+	targetNode := foundName.Target.Owner()
+	docProvider := service.MustGet[DocumentationProvider](s.sc)
```
msujew: Suggestion: The `DocumentationProvider` doesn't have to be set. Use `Get` and return nil if not set.

msujew (commented):
Looking quite good, thanks!

server/documentation_provider.go
```diff
@@ -0,0 +1,111 @@
...
+	lowerBound := 0
+	i := sort.Search(len(doc.Tokens), func(k int) bool {
+		return int(doc.Tokens[k].TextSegment.Indices.Start) >= nodeStart
+	})
+	if i > 0 {
+		lowerBound = int(doc.Tokens[i-1].TextSegment.Indices.End)
+	}
+
+	// Collect comments in [lowerBound, nodeStart) — doc.Comments is sorted by offset,
+	// so results are already in chronological order; no reversal needed.
```
msujew: Minor: The second part of this comment can be dropped. If refers to code that was removed as part of the review.

server/documentation_provider.go
```diff
@@ -0,0 +1,111 @@
...
+		return int(doc.Comments[j].TextSegment.Indices.Start) >= lowerBound
+	})
+	var parts []string
+	for j := first; j < len(doc.Comments); j++ {
+		c := doc.Comments[j]
+		if int(c.TextSegment.Indices.Start) >= nodeStart {
+			break
+		}
+		parts = append(parts, s.trimmer.TrimComment(c.Image))
+	}
```
msujew: Suggestion: One more thing that I just noticed. I think our default implementation should only merge multiple comments if there is no empty line between them. I.e. something like this should get merged:

```go
// one part of the comment
// second part of the comment
func ...
```

This as well:

```go
// one part of the comment
//
// second part of the comment
func ...
```

But **not** this:

```go 
/* We now begin with the next section */

// just single comment
func ...
```

server/documentation_provider_test.go
```diff
@@ -0,0 +1,35 @@
...
+	assert.Equal(t, "foo bar", (&DefaultDocumentationTrimmer{}).TrimComment("/* foo bar */"))
+}
+
+func TestTrimCommentMultilineBlockComment(t *testing.T) {
+	input := "/*\n * line one\n * line two\n */"
+	assert.Equal(t, "line one\nline two", (&DefaultDocumentationTrimmer{}).TrimComment(input))
+}
+
+func TestTrimCommentEmpty(t *testing.T) {
+	trimmer := &DefaultDocumentationTrimmer{}
```
msujew: Suggestion: Move this into some root-level `var trimmer = &DefaultDocumentationTrimmer{}` statement after the imports. That way you don't need to re-instantiate it on every test.

msujew (commented):
Thanks, much better. Just a few suggestions left, see below.

msujew (approved):
Thanks, looks good to me. Please rebase the PR 👍

examples/statemachine/statemachine_hover_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+// createHoverServices builds a service container with both the statemachine language
+// and the default server services (DocumentationProvider, HoverProvider, etc.).
+func createHoverServices() *service.Container {
+	sc := service.NewContainer()
+	SetupServices(sc)
+	server.SetupDefaultServices(sc)
+	sc.Seal()
+	return sc
+}
```
msujew: I just noticed. You can use `CreateLspServices(nil)` instead here.

PR #100, opened 2026-06-08
Replace ATN magic numbers

### LINKED ISSUE #97: [cmd] Generate names for ATN states
Right now, the parser generates numbers into the parser code for error recovery code:

https://github.com/TypeFox/fastbelt/blob/bf01497eab2e26d49822b0e8b1d9ceaae9f8b1af/examples/statemachine/parser_gen.go#L249-L251

We should replace these with generated constants that are (mostly) stable across generation runs.

msujew:
Closes #97

Contains changes to properly generate names for all ATN states. To make it easier to detect "special" ATN states, those receive a double `__`, whereas "normal" atn states (i.e. those generated via assignments, keywords, rule calls, etc.) simply get one `_`. Also ensures that each name is unique.

Also adjusts the ATN generation code to make it more succinct - it's much shorter now.

Additionally removes a few unnecessary ATN state types.


internal/atn/emit.go
Lotes, line 1: The ATN markdown generator could use the names AND the numeric value at the same time. This might help during debugging of the ATN.

examples/statemachine/completion_parser_gen.go
```diff
@@ -38,7 +38,7 @@ func (p *CompletionParser) Parse(tokens []core.Token) *parser.CompletionParseRes
 }
 
 func (p *CompletionParser) ParseStatemachine() {
-	p.cp.EnterRule("Statemachine", 0)
+	p.cp.EnterRule("Statemachine", Statemachine__Start)
```
Lotes, line 41: Thanks for adding the state names. I hope this improves debugging experience when working with the ATN. Like using the ATN markdown with state numbers or state names.

internal/atn/emit.go
msujew: Good point, didn't think about the generated markdown. Should be fixed now 👍

Lotes: Thanks a lot!

Lotes (approved):

spoenemann:
Fixes #97?

PR #101, opened 2026-06-09
Contribution of the 'arithmetics' example language, validation of Interface field names, supplement in parser generation

sailingKieler:
contributes:
* the 'arithmetics' example language, including
  * type hierarchies: interface inheritance/specialization
  * grammar `action`s: switch the type of the current AST object being parsed
    * in combination with cross-reference properties

* extended interface field name validation in the grammar language

* consideration of grammar actions in generation of cross-ref properties → now generates the correct method name of the reference constructor in such cases

* extension of some testing utilities


internal/vscode-extensions/arithmetics/package.json
msujew, line 1: Suggestion: Belongs to `internal/vscode-extensions` instead.

examples/arithmetics/cmd/arithmetics-lsp/main.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2025 TypeFox GmbH
```
msujew: Update the copyright (2026)

internal/generator/type_generator.go
```diff
@@ -115,6 +121,14 @@ func getFieldInfo(field grammar.Field) FieldInfo {
 	}
 }
 
+func getPropAccessorName(name string) string {
+	return strings.ToUpper(name[0:1]) + name[1:]
+}
```
msujew: Discussion: I'd actually rather have a validation that ensures that names are always `PascalCase`. If we don't do this, we need to validate against rather obscure issues:

```go
interface Some {
  name string
  Name string // <-- same generated property, different name
}
```

examples/arithmetics/.gitignore
msujew, line 1: We have a git ignore over at `internal/vscode-extensions`.

examples/arithmetics/server/main.go
```diff
@@ -0,0 +1,32 @@
...
+	lang "typefox.dev/fastbelt/examples/arithmetics"
+	"typefox.dev/fastbelt/server"
+	"typefox.dev/fastbelt/util/service"
+)
+
+// CreateServices creates a service container for the arithmetics language to be used in the language server.
+func CreateServices() *service.Container {
+	sc := service.NewContainer()
+	lang.SetupServices(sc)
+	server.SetupDefaultServices(sc)
```
msujew, line 20: Suggestion: This should also call `server.SetupStdioServices` now.

msujew (commented):
Please add a new launch config for the new language. Thanks!

sailingKieler:
Thanks for the quick review @msujew.

I generated the language structure with the `scaffold` tool.
Do we wanna diverge regarding the location of the `vscode-extension` here?

I'll contribute a few example specs in the arithmetics language, and maybe a few parsing and linking tests.

The more important part for me:
Do you agree with the changes in the `generator` parts?

msujew:
> Do we wanna diverge regarding the location of the vscode-extension here?

Yes, because otherwise we will have all `node_modules` x times in the repository 🙃 

> I'll contribute a few example specs in the arithmetics language, and maybe a few parsing and linking tests.

👍 

> Do you agree with the changes in the generator parts?

No, as indicated in https://github.com/TypeFox/fastbelt/pull/101#discussion_r3378550395.

sailingKieler:
> > Do we wanna diverge regarding the location of the vscode-extension here?
> 
> Yes, because otherwise we will have all `node_modules` x times in the repository 🙃

Fair point. 👍 
> 
> > I'll contribute a few example specs in the arithmetics language, and maybe a few parsing and linking tests.
> 
> 👍
> 
> > Do you agree with the changes in the generator parts?
> 
> No, as indicated in [#101 (comment)](https://github.com/TypeFox/fastbelt/pull/101#discussion_r3378550395).

Overlooked that one.

internal/generator/type_generator.go
sailingKieler: Both options are possible in Go, determining the field visibility outside of the containing package.
Since we have the (generated) getters anyway, we may argue that the fields are always exported and their definition shall express that.

I stumbled across this while copying the grammar over from Langium.

Regarding the validation and your example:
We should have a duplicate name validation, anyway, and forbid stuff like below,
so validating the capitalization wouldn't be a big deal.
```
interface Some {
  Name string
  Name string
}
```

msujew: Right, but I'm not a big fan of messing with the user input. Either emit what the user wants, or reject invalid inputs. We shouldn't go against the conventions of the target platform (which is `PascalCase` for Go).

examples/arithmetics/server/main.go
sailingKieler: Done.

internal/vscode-extensions/arithmetics/package.json
sailingKieler: done.

internal/vscode-extensions/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "clean": "shx rm -rf ./*/dist ./*/*.tsbuildinfo",
-    "build": "npm run build --workspace=fastbelt --workspace=statemachine"
+    "build": "npm run build --workspace=fastbelt --workspace=arithmetics --workspace=statemachine"
```
msujew: Minor: I believe you can simply run `npm run build --workspaces` here.

internal/grammar/validator.go
```diff
@@ -185,6 +188,38 @@ func checkRuleReturnType(rule ParserRule, _ context.Context, accept core.Validat
 
 func (i *InterfaceImpl) Validate(ctx context.Context, _ string, accept core.ValidationAcceptor) {
 	checkInterfaceExtends(i, ctx, accept)
+	checkInterfaceFieldNames(i, accept)
```
msujew: Question: Do we also want to check for inherited names here? Because inherting something like `Name string` and then declaring it yourself will lead to a compiler error in Go.

msujew (commented):
Thanks, looks quite good to me now 👍

internal/grammar/validator.go
sailingKieler: Good point. I'll address that.

internal/vscode-extensions/package.json
sailingKieler: works.

internal/grammar/validator.go
sailingKieler: done.

At that I upgraded the functionality of `ExpectDiagnostic` to work with multiple diagnostics per text range, the `With...` methods now filter them by applying their criteria.

Ready for another review. 🚀

test/doc_fixture.go
```diff
@@ -56,67 +56,96 @@ func (d *Doc) AssertNoParseErrors() *Doc {
 	return d
 }
 
+// AssertNoLinkingErrors fails the test if the document has no references at all,
+// or if any reference failed to resolve.
```
msujew: Suggestion: This documentation is incorrect? As far as I can tell, the test will pass if `d.Document.References` is empty.

test/doc_fixture.go
```diff
@@ -125,13 +154,34 @@ func (d *Doc) ExpectDiagnostic(label string) *DiagnosticExpectation {
...
-	return nil
+	if len(matched) == 0 {
+		d.fixture.t.Fatalf("fbtest: no diagnostic at label %q", label)
+		return nil
+	}
+	return &DiagnosticExpectation{t: d.fixture.t, Diagnostics: matched}
+}
+
+// ExpectNoDiagnostic fails the test if at least one diagnostic has the given
+// severity and a message containing msgSubstring.
```
msujew: Suggestion: This comment is also incorrect.

test/doc_fixture.go
```diff
@@ -125,13 +154,34 @@ func (d *Doc) ExpectDiagnostic(label string) *DiagnosticExpectation {
...
+func (d *Doc) ExpectNoDiagnostic(label string) *Doc {
```
msujew: Minor: Rename to `AssertNoDiagnostic`. The idea behind the `Expect*` methods is to return a `Expectation` struct which can then be used further in the tests, whereas the `Assert*` methods are designed to simply assert some kind of document state.

msujew (commented):
Thanks, looks quite good to me. Just a few minor comments, see below 👍

test/doc_fixture.go
sailingKieler: Right, doc seems like a leftover from a previous state. Fixed.

test/doc_fixture.go
sailingKieler: Fixed.

test/doc_fixture.go
sailingKieler: Done.

sailingKieler:
Final remarks are treated.

msujew (approved):
Looks good to me 👍

PR #102, opened 2026-06-09
Run benchmark in CI

msujew:
Adds a GitHub action that runs the whole benchmark suite in the CI, prints a summary of the benchmarks and shows a warning if any of them degrade by more than 50%. Note that a degradation does not impact the failure-state of the CI - it is still possible to merge the PR.

Removes the internal benchmarking script in favor of using longer benchmarking numbers.


msujew:
FYI, I disabled the comment that always appears for now. Devs can always take a look at the job summary, see [here](https://github.com/TypeFox/fastbelt/actions/runs/27210037939) for an example.

Lotes:
Is it actually possible to run this in an own isolated environment, so that the delta numbers are near zero? Because you just changed the GH action and the ratio is between +/-20%.

I guess this is for detecting situations where performance got really worse by accident ^^*.

msujew:
@Lotes Would be great, but there no "please just give me a reproducible environment" button in GitHub. I had to turn up the threshold from 120% to 200%, otherwise noise got through the benchmark. Quite unfortunate.

msujew:
I'll continue working on this - I believe there is some room for improvement. We can probably run the benchmarks multiple times, post-process the results before passing them to the CI action for summarization.

msujew:
The latest changes seemed to have the desired effect. Benchmarks are roughly within ~10% margin of error. This is good enough for a CI. So this is good enough for a review.

Lotes (approved):
Thanks, looks good. Have you tried out higher times like 10 or 20 s?
Why not higher :-) ? I guess the total time becomes to large then, or?

msujew:
> Why not higher :-) ? I guess the total time becomes to large then, or?

Yes, we will probably add quite a few more benchmarks in the future. I don't want to extend the CI too much. We can think about it once we're a bit more "settled in" so to speak.

PR #103, opened 2026-06-09
Proposal: Eager output channel setup in VS Code extensions & server initialization heartbeat log msgs

sailingKieler:
For the sake of having LS output channels in VS Code's output view while firing up the VSC extension, see screenshot,
this change adds explicit channel creations to the existing VSC extensions + the extension template.
(By default VS Code creates them lazily on the first output.)

Proposal: This change adds some heartbeat log messages before and after the LS initialization.
We might make that configurable, but (in development) it's often nice to see it's alive and properly initialized.

image

### DIFF
--- a/internal/scaffold/templates/vscode-extension/src/extension.ts.tmpl
+++ b/internal/scaffold/templates/vscode-extension/src/extension.ts.tmpl
@@ -2,7 +2,7 @@
 // This program and the accompanying materials are made available under the
 // terms of the MIT License, which is available in the project root.
 
-import type * as vscode from 'vscode';
+import * as vscode from 'vscode';
 import type { LanguageClientOptions, ServerOptions } from 'vscode-languageclient/node.js';
 import * as path from 'node:path';
 import { LanguageClient, TransportKind } from 'vscode-languageclient/node.js';
@@ -28,7 +28,12 @@ async function startLanguageClient(context: vscode.ExtensionContext): Promise<La
     };
 
     const clientOptions: LanguageClientOptions = {
-        documentSelector: [{ scheme: 'file', language: '{{.LanguageID}}' }]
+        documentSelector: [{ scheme: 'file', language: '{{.LanguageID}}' }],
+        // explicitly create an output channel enabling monitoring the LS logs from the beginning in vs code's 'output' view
+        // (by default vs code creates such channels lazily on the first message)
+        outputChannel: vscode.window.createOutputChannel(
+            '{{.LanguageLabel}} Language Server'
+        )
     };
     const lc = new LanguageClient(
         '{{.LanguageID}}',
--- a/internal/vscode-extensions/fastbelt/src/extension.ts
+++ b/internal/vscode-extensions/fastbelt/src/extension.ts
@@ -2,7 +2,7 @@
 // This program and the accompanying materials are made available under the
 // terms of the MIT License, which is available in the project root.
 
-import type * as vscode from 'vscode';
+import * as vscode from 'vscode';
 import type { LanguageClientOptions, ServerOptions } from 'vscode-languageclient/node.js';
 import * as path from 'node:path';
 import { LanguageClient, TransportKind } from 'vscode-languageclient/node.js';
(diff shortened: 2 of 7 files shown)


internal/scaffold/templates/vscode-extension/src/extension.ts.tmpl
```diff
@@ -28,7 +28,12 @@ async function startLanguageClient(context: vscode.ExtensionContext): Promise<La
     };
 
     const clientOptions: LanguageClientOptions = {
-        documentSelector: [{ scheme: 'file', language: '{{.LanguageID}}' }]
+        documentSelector: [{ scheme: 'file', language: '{{.LanguageID}}' }],
+        // explicitly create an output channel enabling monitoring the LS logs from the beginning in vs code's 'output' view
+        // (by default vs code creates such channels lazily on the first message)
+        outputChannel: vscode.window.createOutputChannel(
+            '{{.LanguageLabel}} Language Server'
```
msujew, line 35: I remember that in VS Code we just use `LanguageLabel` - or rather that's what the `vscode-languageserver` has been doing. IMO, we should do the same.

internal/vscode-extensions/fastbelt/src/extension.ts
```diff
@@ -32,7 +32,10 @@ async function startLanguageClient(context: vscode.ExtensionContext): Promise<La
 
     // Options to control the language client
     const clientOptions: LanguageClientOptions = {
-        documentSelector: [{ scheme: 'file', language: 'fastbelt' }]
+        documentSelector: [{ scheme: 'file', language: 'fastbelt' }],
+        outputChannel: vscode.window.createOutputChannel(
+            'Fastbelt Grammar Language Server'
```
msujew, line 37: Suggestion: For users, it's not relevant that the info comes from a language server - the important part is that it comes from *the language*. I think just `Fastbelt` is enough here. Same with the statemachine language.

sailingKieler: Well, I have a different opinion here.
We're speaking about the console of logs of implementation internals that is consulted by interested users only. I wouldn't judge what's relevant for them.
Since we have the opportunity of providing sensible name I'm in favor of using it.

Other implementations also distinguish LS log channels and other extension related ones.

image

Typescript doesn't seem to do it, or the TS LS  just doesn't log anything from its internals.

Besides, my inner monk can't deal with the message of _Statemachine_ or _Arithmetics_ or _Fastbelt_ did crash with a seg fault.
 → Which Statemachine?
 → Or which part of it?

msujew: Ok, I can agree with that. Since adopters can freely change the name of the output channel as well, I'm fine with it. Thanks for the input 👍

server/server.go
```diff
@@ -69,7 +70,9 @@ func (s *DefaultLanguageServer) Initialize(ctx context.Context, params *lsp.Para
 }
 
 func (s *DefaultLanguageServer) Initialized(ctx context.Context, params *lsp.InitializedParams) error {
+	log.Print("LS initializing...")
 	if initializer, err := service.Get[workspace.Initializer](s.sc); err == nil {
+		defer log.Println("done.")
```
msujew: Minor: Move the `LS initializing` message into the branch? Because it's not actually initializing if no `Initializer` is registered. Also, maybe we can improve the message to make it clear that these are connected.
```suggestion
if initializer, err := service.Get[workspace.Initializer](s.sc); err == nil {
        log.Print("LS initializing...")
		defer log.Println("LS initialization done.")
```

msujew (approved):
Thanks, looks good to me. Just a minor comment, see below.

internal/vscode-extensions/fastbelt/src/extension.ts
sailingKieler: Added the channel for the 'Arithmetics' language and the 'Statemachine' web variant, too.

server/server.go
sailingKieler: You're right, I misunderstood the `.Get[workspace.Initializer](s.sc)` part.
Updated this part, too.

What do you think about the new current message?

msujew (approved):
Thanks, looks good to me 👍

PR #104, opened 2026-06-11
Adaptive Lookahead

### LINKED ISSUE #74: Lookahead problems with (composite?) rules
when porting from langium we have a grammar excerpt like

```
grammar Repro;

interface Root { Value composite }

composite ID: RawId | EscapedId;
composite QualifiedName: ID ("." ID)*;   // greedy — consumes all dot-separated IDs
composite PATH: ID ("/" ID)* ":" ID | ID;
composite QualifiedPATH: QualifiedName "::" PATH | PATH;

token RawId:     /[_a-zA-Z][\w_-]*/;
token EscapedId: /`[^`]*`/;
hidden token WS: /[ \t\r\n]+/;

Root returns Root:
    Value=QualifiedPATH;
```

and expect to be able to parse

| Input | Expected `QualifiedPATH` result | Notes |
|---|---|---|
| `myId` | `myId` | plain PATH (single ID) |
| `some/path:node` | `some/path:node` | PATH with slash+colon |
| `pkg::node` | `pkg::node` | single-segment qualified name |
| `a.b.c::node` | `a.b.c::node` | multi-segment qualified name |

but

`myId` 

does not parse

extended to

```
EnumMemberValue returns EnumMemberValue:
    "enum-member" "<" EnumRef=QualifiedPATH "." MemberName=ID ">";
```

```enum-member<EntityType.LegalEntity>``` also does not parse

### PR DESCRIPTION
**Pull request by msujew**
Implements the ALL(*) algorithm (from ANTLR4) in Fastbelt:

1. Generates a new `<language>ParserLookahead` interface and struct to make the lookahead overridable (has a slightly negative impact on performance).
2. Adaptive LL(*) is called based on need - LL1 decisions are generated if not necessary.
3. Adds a new lookahead language that is used to test the whole suite of lookahead features.
4. Implements both SLL (strong LL, known from basic recursive descent parsers) and full-context LL (partly implemented in Chevrotain, required for certain parsing decisions in our test language).
5. Fixes a bug in the ATN generation related to cardinality of rule calls in cross references.
6. Improves the error reporting for alternatives (now says which tokens were expected).

### CODE REVIEW

File: util/collections/bitset.go [hunk 1]
```diff
@@ -70,72 +49,59 @@ func NewBitset() *BitSet {
...
-			// Special case if bitset is still empty
-			b.offset = index
-			b.set = make([]uint64, 1)
-			delta = 0
-		} else {
-			intermediate := make([]uint64, delta+1)
-			copy(intermediate, b.set)
-			b.set = intermediate
-		}
+	w := i >> 6
```
Lotes, line 57: `x >> 6` is equal to `x / 64`...
`x << 6` is equal to `x * 64`...
Thinking about these magic numbers, if we can use constants here. But what would be a good name? Could imagine that the compiler is smart enough to optimize on its own.

util/collections/bitset.go
```diff
@@ -70,72 +49,59 @@ func NewBitset() *BitSet {
...
+	for w >= len(b.words) {
+		b.words = append(b.words, 0)
 	}
-	b.set[delta] |= 1 << offset
+	b.words[w] |= 1 << (uint(i) & 63)
```
Lotes: `(uint(i) & 63)` could be `i % 64`. Magic number.
I think `/` and `%` are easier to grasp. WDYT?

util/collections/bitset.go
Lotes: There are multiple places for these examples in this file here.

parser/prediction_context.go
```diff
@@ -0,0 +1,429 @@
...
+	parent      *predictionContext
+	returnState int
+
+	// array
+	parents      []*predictionContext
+	returnStates []int
+}
+
+const (
+	pcEmpty = iota
```
Lotes: Was it not possible "to type" this enum using a name? With the effect that only these 3 values can be assigned to the prediction contexts type...

internal/languages/lookahead/lookahead.fb
Lotes, line 1: Is it possible or of value having a test case with two identical alternatives? The lookahead would choose the first one and/or print an ambiguity report.

internal/generator/parser_generator.go
```diff
@@ -13,52 +13,372 @@ import (
...
+	_, loopIsAdaptive := ctx.atnData.loopAdaptive[alts]
+	_, orIsAdaptive := ctx.atnData.orAdaptive[alts]
+	return !loopIsAdaptive && !orIsAdaptive
+}
+
+// canCombineAlternatives is canCombineAltsWithLoop restricted to Alternatives
+// nodes whose own cardinality is non-One (so the loop guard is on the Alternatives
+// itself, not on a wrapping Assignment or Group).
+func (ctx *ParserGeneratorContext) canCombineAlternatives(alts grammar.Alternatives) bool {
+	return alts.Cardinality() != CardinalityOne && ctx.canCombineAltsWithLoop(alts)
```
Lotes, line 138: It feels weird to compare to strings for cardinality. Using constants is a step in the right direction. Why don't we use enums?
Is it faster to compare two runes instead of two int32s?

internal/generator/parser_generator.go
```diff
@@ -13,52 +13,372 @@ import (
...
+			break
+		}
+	}
+
+	// Build the maps once; generateLL1Lookahead needs them per table.
+	varNameToId := buildVarNameToId(tokenTypes)
+	groupMembers := buildGroupVarNameToMembers(grammr)
+
+	node := codegen.NewNode()
+	node.AppendLine("// Code generated by typefox.dev/fastbelt/cmd/fastbelt. DO NOT EDIT.")
```
Lotes: This line might be good kept in a shared constant

internal/generator/parser_generator.go
```diff
@@ -219,7 +511,8 @@ func GenerateCompletionParser(grammr grammar.Grammar, packageName string, tokenT
 	node.Indent(func(n codegen.Node) {
 		n.AppendLine("messages := service.MustGet[parser.ErrorMessageProvider](p.sc)")
 		n.AppendLine("recovery := service.MustGet[parser.ErrorRecoveryStrategy](p.sc)")
-		n.AppendLine("cp := &CompletionParser{sc: p.sc, atn: p.atn}")
+		n.AppendLine("lookahead := service.MustGet[", grammr.Name(), "ParserLookahead](p.sc)")
```
Lotes, line 508: I read `grammr.Name(), "ParserLookahead"` > 3 times in this file. You could save that in a constant.

internal/grammar/parser_lookahead_gen.go
```diff
@@ -0,0 +1,267 @@
...
+)
+
+const (
+	DecisionElementAlternatives = 16
+	DecisionGrammarAlternatives = 0
+)
+
+var ActionOperatorAlternatives = parser.LL1Lookahead{
+	Types:  []*core.TokenType{Keyword_PlusEquals, Keyword_Equals},
+	Lookup: []int{5: 1, 10: 2},
```
Lotes, line 17: Inside the lookup: what is the meaning of the index? I guess it is a state number. We could reuse the state number constants here to avoid magic numbering.
Using the constants would help understanding the code better, since there is a connection via the name that points to the concept of ATN states or what ever.

(But I actually did not do it for NFA state lookups, so it is up to you. WDYT?)

internal/languages/lookahead/parser_test.go
```diff
@@ -0,0 +1,204 @@
...
+	doc, err := fastbelt.NewDocumentFromString("file:///workspace/test.la", "lookahead", content)
+	if err != nil {
+		b.Fatal(err)
+	}
+	doc.Tokens = tokens
+	b.SetBytes(int64(len(content)))
+	b.ResetTimer()
+	for b.Loop() {
+		result := parserService.Parse(doc)
+		doc.Root = result.Node
```
Lotes: What is the purpose of this line?

internal/languages/token_groups/atn_gen.go
Lotes, line 1: Can we generate the ATN markdown by default? What speaks against this?

parser/prediction_context.go
```diff
@@ -0,0 +1,429 @@
...
+type predictionContext struct {
+	cachedHash uint64
+	pcType     int
+
+	// singleton
+	parent      *predictionContext
+	returnState int
+
+	// array
+	parents      []*predictionContext
```
Lotes, line 19: 1. You could merge the fields to use only parents and returnStates.

2. How to imagine a prediction context with multiple parents per frame? What relation does the chain of parent have then? Hard to grasp.

parser/parser_atn_simulator.go
```diff
@@ -0,0 +1,416 @@
...
+	PredictionModeSLL PredictionMode = iota
+	// PredictionModeLL allows the current parser context to be used when making a prediction
+	// It is more expensive than [PredictionModeSLL], but is guaranteed to produce correct
+	// predictions for all syntactically valid inputs.
+	PredictionModeLL
+)
+
+// FullContext reports whether this prediction mode performs full-context LL
+// prediction (true) or uses Strong LL (SLL, false).
+func (m PredictionMode) FullContext() bool {
```
Lotes, line 55: `IsRunningFullContextPrediction`?

parser/dfa.go
```diff
@@ -0,0 +1,76 @@
...
+//
+// The dfa is shared across concurrent parses (it hangs off the shared
+// RuntimeATN), so all mutation of s0, the state set, and edges is guarded by
+// mu.
+type dfa struct {
+	decision      int
+	atnStartState *RuntimeATNState
+
+	mu     sync.RWMutex
+	s0     *dfaState
```
Lotes: Name it `startState`?

parser/parser_atn_simulator.go
```diff
@@ -0,0 +1,416 @@
...
+	if s0 == nil {
+		s0Closure := s.computeStartState(dfa.atnStartState, emptyPredictionContext(), false)
+		s0 = dfa.addState(newDFAState(-1, s0Closure))
+		dfa.setS0(s0)
+	}
+	return s.execATN(dfa, s0, mode, outerContext)
+}
+
+func (s *parserATNSimulator) execATN(dfa *dfa, s0 *dfaState, mode PredictionMode, outerContext *predictionContext) (int, *PredictionFailure) {
+	previousD := s0
```
Lotes: D == Donuts? Dinosaurs? Dont know? :D
`previousDFAState`?

parser/parser_atn_simulator.go
```diff
@@ -0,0 +1,416 @@
...
+	offset := 1
+	t := s.parserState.LA(offset)
```
Lotes: I would prefer longer names, like `token`, `sim`, `nextState`.
One has to guess or read more code to understand it.

parser/parser_atn_simulator.go
```diff
@@ -0,0 +1,416 @@
...
+		}
+	}
+	sort.Slice(expected, func(i, j int) bool {
+		return expected[i].Name < expected[j].Name
+	})
+	return &PredictionFailure{Token: token, Expected: expected}
+}
+
+func (s *parserATNSimulator) computeTargetState(dfa *dfa, previousD *dfaState, mode PredictionMode, t *core.TokenType) *dfaState {
+	reach := s.computeReachSet(previousD.configs, t, mode.FullContext())
```
Lotes, line 150: May name it `reach(ATN)ConfigSet` to make the type readable

Lotes: Same for other occurences

parser/parser_atn_simulator.go
```diff
@@ -0,0 +1,416 @@
...
+		reach.conflictingAlts = s.getConflictingAlts(reach)
+		d.requiresFullContext = true
+		d.isAcceptState = true
+		d.prediction = reach.conflictingAlts.Min()
+	}
+	return dfa.addEdge(previousD, t, d)
+}
+
+func (s *parserATNSimulator) execATNWithFullContext(s0 *atnConfigSet) (int, *PredictionFailure) {
+	const fullCtx = true
```
Lotes, line 170: inline the variable

parser/dfa.go
```diff
@@ -0,0 +1,76 @@
...
+// adaptive prediction.
+//
+// The dfa is shared across concurrent parses (it hangs off the shared
+// RuntimeATN), so all mutation of s0, the state set, and edges is guarded by
+// mu.
+type dfa struct {
+	decision      int
+	atnStartState *RuntimeATNState
+
+	mu     sync.RWMutex
```
Lotes, line 23: Puh, `mutex` is already abbreviated enough. `mu` is too short in my eyes.
Question marks will arise over the readers head. Danger. Danger. :D

parser/dfa_state.go
```diff
@@ -0,0 +1,58 @@
...
+func (d *dfaState) getEdge(t *core.TokenType) *dfaState {
+	i := t.Id + 1
+	if i < 0 || i >= len(d.edges) {
+		return nil
+	}
+	return d.edges[i]
+}
+
+func (d *dfaState) setEdge(t *core.TokenType, target *dfaState) {
+	i := t.Id + 1
```
Lotes: Using one-letter-variables makes often reading harder.

parser/atn_config_set.go
```diff
@@ -0,0 +1,218 @@
...
+type configKey struct {
+	state *RuntimeATNState
+	alt   int
+}
+
+// atnConfigSet is a set of ATNConfig that merges configs sharing (state, alt)
+// by combining their prediction contexts, and tracks the bookkeeping the
+// predictor needs (uniqueAlt, conflictingAlts, dipsIntoOuterContext, fullCtx).
+type atnConfigSet struct {
+	configs []*atnConfig
```
Lotes, line 29: A config set contains multiple configs. What do these configs have in common? Why are they united into a set? How can I imagine?

Lotes (commented):
Thanks! I have some questions and code quality proposals.
I have problems in understanding the ATN config set, see question below. Would be nice to understand what the structure/the relations of the members in theses sets are :)...

parser/dfa.go
msujew: Standard for Go, see also [here](https://go.dev/tour/concurrency/9).

parser/atn_config_set.go
msujew: In essence, the ALL(*) algorithm performs the NFA subset construction algorithm. Each `atnConfig` is one ATN/NFA state, whereas each `atnConfigSet` is the set of "current" NFA states, i.e. represents one DFA state. Each `atnConfigSet` is attached to one `dfaState` struct instance.

parser/parser_atn_simulator.go
msujew: I actually like the constant here. It's makes the downstream code (where it is used) a bit clearer. Note that the compiler inlines it anyway.

util/collections/bitset.go
msujew: Good point, the compiler will optimize this. Added a `wordSize` constant.

internal/generator/parser_generator.go
msujew: `Cardinality()` returns a string, and since `grammar.Altneratives` is a part of the generated types, we cannot easily change the parsed string to an int.

internal/grammar/parser_lookahead_gen.go
msujew: These are actually token type IDs (no ATN usage, due to LL(1)). Since the numbers in there are magic numbers anyway (the alternative indices), I would also keep the array indices for previty. Otherwise we get `Keyword_PlusEquals_Idx: 1`, etc. These can become quite long for more complex alternatives.

internal/languages/token_groups/atn_gen.go
msujew: What for?

parser/parser_atn_simulator.go
msujew: I would err on the side of brevity. I like `if mode.FullContext()` more than `if mode.IsRunningFullContextPrediction()`.

parser/parser_atn_simulator.go
msujew: I would keep it short. Not only for brevity, but also because it makes it clearer to people who want to compare the implementation with the original code (from ANTLR, where it has the same name).

internal/grammar/parser_lookahead_gen.go
Lotes: I have a mixed opinion here. Let's keep it like it is.

internal/languages/token_groups/atn_gen.go
Lotes: If you want to debug a specific ATN. Most of the time you will not need it, sure.
But one could argue, that you do not have to search the "generate" command again.

parser/parser_atn_simulator.go
Lotes: The length is debatable. I just missed a verb pointing to the boolean nature of the member.
If I read FullContext, it could be also an object and the condition a non-null check.

parser/parser_atn_simulator.go
Lotes: Yes, true. Can live with it. After reading more methods it becomes clear what it is.

parser/parser_atn_simulator.go
Lotes: It is just weird, because it is the only usage AND constant.

A options object would be nicer actually. But only in JS... in Go that looks like too much.

Lotes: For readibility you could rename it to `enableFullCtx`, but I do not insist on it.

internal/languages/lookahead/lookahead.fb
```diff
@@ -52,6 +52,13 @@ H returns Obj: "h" (
     | Node=QualifiedNameRecursive Value="world"
 );
 
+I returns Obj: "i" (
+    // Ambiguous alternatives
```
Lotes, line 56: Thx

parser/parser_atn_simulator.go
msujew: I've renamed it to `IsFullContext()`

internal/generator/parser_generator.go
msujew: This applies to most other services as well. I'll keep it as it is for now. I don't think it makes sense to generate everything into a constant that is used more than once throughout the generators.

internal/languages/token_groups/atn_gen.go
Lotes: It is probably enough to do it for the main grammar. One problem that could arise is when we would disable it for all grammar, we might forget that the `ATN.md` exists (also because there are no error messages by its nature). We almost forgot it already once.

I like the MD file: instead of having to deal with pure numbers or state names, not knowing who is connected to whom, you have these nice diagrams that displays the relations clearly.

internal/generator/parser_generator.go
Lotes: OK, it was an observation. I do not think we will rename this so fast. But that is the point: if we need to, we need to find every occurrence by hand, instead of having one variable.

I also had blacklists in mind. I do not know when this becomes relevant: avoid generated name collisions with user variables. (Can be a really "minor to nothing" thing)

internal/languages/token_groups/atn_gen.go
msujew: I would keep it for the main grammars. It's not really useful for the test languages - especially because GitHub doesn't understand that it's a generated file.

internal/generator/parser_generator.go
msujew: > I also had blacklists in mind. I do not know when this becomes relevant: avoid generated name collisions with user variables. 

That's a good point. We should probably resolve this before the release.

Lotes (approved):
Conceptually, it is not so hard to understand. The details look a bit different: but there are a lot of comments, nice.
If you want more review comments, ask someone else as well.

From my high level perspective I am fine with the code, thanks :-).

PR #105, opened 2026-06-15
Improve SlogHandler implementation

msujew:
Performs some minor improvements to the `SlogHandler` (now public) to ensure that we forward the correct LSP log level. Previously, it only forwarded `lsp.Info`, completely disregarding the actual log level.

Also adds documentation.

### DIFF
--- a/server/slog_handler.go
+++ b/server/slog_handler.go
@@ -7,47 +7,55 @@ package server
 import (
 	"context"
 	"log/slog"
-	"strings"
 
 	"typefox.dev/fastbelt/util/service"
 	"typefox.dev/lsp"
 )
 
-type slogHandler struct {
+// SlogHandler is a [slog.Handler] that sends log messages to the
+// LSP client via the connection. Note that this handler does not
+// take attributes or groups into account when writing the message
+// to the language client.
+type SlogHandler struct {
 	sc *service.Container
 }
 
 // NewSlogHandler creates a new slog handler that writes to the LSP connection.
 func NewSlogHandler(sc *service.Container) slog.Handler {
-	return &slogHandler{sc: sc}
+	return &SlogHandler{sc: sc}
 }
 
-func (h *slogHandler) Enabled(ctx context.Context, level slog.Level) bool {
+func (h *SlogHandler) Enabled(ctx context.Context, level slog.Level) bool {
 	// Enable all log levels. You can customize this to filter based on level or context.
 	return true
 }
 
-func (h *slogHandler) Handle(ctx context.Context, record slog.Record) error {
-	sb := strings.Builder{}
-	sb.WriteString(record.Time.Format("15:04:05.000"))
-	sb.WriteString(" [")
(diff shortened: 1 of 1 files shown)


server/slog_handler.go
```diff
@@ -7,47 +7,55 @@ package server
...
 	"typefox.dev/fastbelt/util/service"
 	"typefox.dev/lsp"
 )
 
-type slogHandler struct {
+// SlogHandler is a [slog.Handler] that sends log messages to the
+// LSP client via the connection. Note that this handler does not
+// take attributes or groups into account when writing the message
+// to the language client.
+type SlogHandler struct {
```
Lotes, line 19: What is a Slog? And attributes and groups in this context? Weird terminology.
I looked it up with AI...

Why dont we support groups and attributes? Seem to be useful.

msujew: See [slog](https://go.dev/blog/slog). Groups and attributes are mostly useful for something like a backend server that sends its logs to some log aggregration system like [`loki`](https://grafana.com/oss/loki/). I'm not even sure how we would actually send the group/attribute data across to the language client for display. In a json format? Note that the registered `slog.Handler` is actually used by the `log` package for simple log printing via `log.Printf(...)`.

server/slog_handler.go
```diff
@@ -7,47 +7,55 @@ package server
...
-	sb := strings.Builder{}
-	sb.WriteString(record.Time.Format("15:04:05.000"))
-	sb.WriteString(" [")
-	sb.WriteString(record.Level.String())
-	sb.WriteString("] ")
-	sb.WriteString(record.Message)
-	return h.write(ctx, lsp.Info, sb.String())
+func (h *SlogHandler) Handle(ctx context.Context, record slog.Record) error {
+	var lspLevel lsp.MessageType
+	switch {
```
Lotes, line 35: I have the feeling this could be written simpler using "less or equal" and the pairs of log levels that correspond to each other:

```go
switch {
	case record.Level <= slog.LevelInfo:  //same (info)
		lspLevel = lsp.Info                 //same (info)
	case record.Level <= slog.LevelWarn:
		lspLevel = lsp.Warning
	case record.Level <= slog.LevelError:
		lspLevel = lsp.Error
	default:
		lspLevel = lsp.Log
	}
```

But I doubt that this means the same ^^* ...

Lotes (approved):
Just a minor thought about the switch. Otherwise looks good :). Thanks.

server/slog_handler.go
msujew: Yeah, unfortunately not possible with how `slog.Level*` values are written. See the source:

```go
const (
	LevelDebug Level = -4
	LevelInfo  Level = 0
	LevelWarn  Level = 4
	LevelError Level = 8
)
```

Note that a `Level` of `2` would be `Info+2` (i.e. 2 levels higher than `Info`, but not quite worth a `Warn`). That's why the switch-case has to look as it does.

PR #106, opened 2026-06-15
Remove root node folding range

msujew:
I've noticed that our root node currently also receives a folding range that effectively covers the whole document. This doesn't really make too much sense. This change simply excludes the root node from the folding ranges.

Can be tested using the statemachine or fastbelt grammar extension - writing dedicated tests for the absence of a folding range doesn't work too well, but manual testing should cover this well enough.


ssmifi (approved):
Makes sense, thx!

PR #107, opened 2026-06-15
Improve offset computation for LSP services

msujew:
Fixes an issue in the LSP services: If you were running an LSP operation at the end of a token, it would attempt to compute the result for the next token (if that tokens starts at the position where the previous token ends).

Since we don't know which token is really semantically relevant, we actually need to adjust our `NameFinder#Find` method to accept both tokens in such a case. It will then try to identify which one is more relevant for the request. As the `FoundName` type already returns the `Source` range, callers will always know whether the `first` or `second` token is the revelant one.

In Langium we solved this issue by taking the first token, if it ends on a unicode letter. While this works fairly well, it's not 100% sound.

We could change the name of `SearchOffset2` to something else - but it's in line with Go convention, see `iter.Seq2` which is `iter.Seq`, but with a tuple.

### DIFF
--- a/examples/statemachine/statemachine_definition_test.go
+++ b/examples/statemachine/statemachine_definition_test.go
@@ -51,3 +51,24 @@ func TestDefinitionOnReference(t *testing.T) {
 
 	doc.AssertDefinition("flick")
 }
+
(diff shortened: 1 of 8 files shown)


spoenemann (approved):
Sounds reasonable! 👍

PR #108, opened 2026-06-17
Package documentation: linking, workspace

spoenemann:
I added more detailed package documentations to `linking` and `workspace` because these are quite complex and may be hard to grasp if you haven't worked with Fastbelt or Langium before.


linking/doc.go
```diff
@@ -0,0 +1,90 @@
...
+// Package linking resolves cross-references in parsed documents and builds the
+// symbol tables and scopes that reference resolution depends on. The
+// [typefox.dev/fastbelt/workspace] builder orchestrates these steps during
+// document processing.
+//
+// # Services
+//
+// [SetupDefaultServices] registers the framework services:
+//
+//   - [SymbolExporter] — symbols this document exposes to other documents
```
msujew, line 14: Question: Do we want to have a pass through the docs to replace em-dash with simple `-` before the release? I'm not a fan of non-ascii symbols in code (also in documentation). I know that we have quite a few of those in our current codebase. But I'm open for other perspectives on this.

msujew (approved):
Thanks, looks good to me 👍

linking/doc.go
spoenemann: It may be confusing in the source code, but it looks good in the generated documentation (preview with `go doc -http`):

Screenshot 2026-06-18 at 09 48 40

PR #109, opened 2026-06-18
Implemented detection of affected documents via reference descriptions

spoenemann:
The approach based on reference descriptions is the same as in Langium.

What's different compared to Langium is that we still don't do any initial workspace build. The first build is triggered when a file is opened. Adopters can change that if they prefer immediate feedback without open files (combined with a broad activation event of the vscode extension).

I also created a Set type, as I find the `map[T]struct{}` pattern hard to read.


util/extiter/extiter.go
Lotes, line 1: What does extiter mean? Cannot find a translation.

msujew: It's an **ext**ension to ther `iter` package.

spoenemann: **Ext**raordinarily awesome **iter**ation!!!

workspace/document_updater.go
```diff
@@ -73,3 +85,45 @@ func (s *DefaultDocumentUpdater) Update(ctx context.Context, changed []textdoc.H
...
+	return isAffected(doc, changedURIs)
+}
+
+// isAffected reports whether doc contains a cross-document reference to any of the
+// documents identified by changedURIs. URIs are compared using
+// [core.URI.StringUnencoded].
+//
+// References to nodes within doc itself are ignored: such local references cannot
+// be invalidated by a change to another document.
+func isAffected(doc *core.Document, changedURIs collections.Set[string]) bool {
```
msujew: Suggestion: Can we move this into a dedicated service interface? It's not uncommon to modify this behavior and doing so would require adopters to override the whole `Update` method right now.

util/collections/set.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, value := range values {
+		s[value] = struct{}{}
+	}
+	return s
+}
+
+// Add inserts value into the set. Adding a value already present is a no-op.
+func (s Set[T]) Add(value T) {
+	s[value] = struct{}{}
+}
```
msujew: Suggestion: Add `bool` return type with similar semantics to other languages. That way you can cut down on the `Has`+`Add` usage. See also (from the ALL(*) PR):

https://github.com/TypeFox/fastbelt/blob/e675ad0eac140b48749e7ba8a02b801f9896a8c7/util/collections/bucket_map.go#L125-L128

util/collections/set.go
```diff
@@ -0,0 +1,59 @@
...
+// Has reports whether value is a member of the set.
+func (s Set[T]) Has(value T) bool {
+	_, ok := s[value]
+	return ok
+}
+
+// Remove deletes value from the set. Removing an absent value is a no-op.
+func (s Set[T]) Remove(value T) {
+	delete(s, value)
+}
```
msujew: Same here.

msujew (commented):
Thanks, that should do well. Just a few improvements on the API.

workspace/document_updater.go
```diff
@@ -73,3 +86,62 @@ func (s *DefaultDocumentUpdater) Update(ctx context.Context, changed []textdoc.H
...
+//
+// A document is relinked when it has an unresolved reference, which might resolve
+// against a symbol introduced by the change, or when it references any of the
+// changed documents (see [DocumentChangeImpact.Affected]).
+func shouldRelink(doc *core.Document, changedURIs collections.Set[string], changeImpact DocumentChangeImpact) bool {
+	for _, ref := range doc.References {
+		if ref.Error() != nil {
+			return true
+		}
+	}
```
msujew: Suggestion: Now that I think about this, can we also move this function into the `DocumentChangeImpact`? Theoretically, not all documents with reference errors need to be rebuilt. In a lot of languages, this can also be inferred from the import graph. If the import graph is disjointed, there's no need to rebuild the document, even in the presence of reference errors.

msujew (approved):
Thanks. One more comment - It would be great to include it, since it's part of the API. Otherwise, looks good to me 👍

PR #110, opened 2026-06-18
Implement file system abstraction

msujew:
Adds a file system abstraction that can be used to implement different kinds of file system (for example - via the vscode API for web extensions).

The default implementation uses the disk as its source of data. Also adds a virtual file system implementation that can be used for testing.

### DIFF
--- a/uri.go
+++ b/uri.go
@@ -36,6 +36,8 @@ type URI interface {
 	// FilePath returns the URI as a file path, which is the format used by the local file system.
 	FilePath() string
 	// JoinPath joins the given path segments to the URI's path and returns a new URI.
+	// It will automatically attempt to normalize the resulting path by resolving "." and ".." segments,
+	// and by collapsing multiple slashes.
 	JoinPath(segments ...string) URI
 	// DocumentURI converts the URI to a lsp.DocumentURI, which is the format used by the LSP library.
 	DocumentURI() lsp.DocumentURI
@@ -228,7 +230,6 @@ func (u *uri) FilePath() string {
 		// UNC path: \\server\share\file.txt
 		sb.WriteString(`\\`)
 		sb.WriteString(u.Authority())
-		isWindows = true
 	}
 	if isWindows && !withAuthority && strings.HasPrefix(path, "/") {
 		// Omit first slash for normal (non-UNC) windows file paths
@@ -246,13 +247,40 @@ func (u *uri) JoinPath(segments ...string) URI {
 	if len(segments) == 0 {
 		return u
 	}
-	joined := strings.Join(segments, "/")
-	newPath := u.Path()
-	if !strings.HasSuffix(u.Path(), "/") {
-		newPath += "/"
+	// An empty path is treated as the root, so the result stays absolute.
+	absolute := u.Path() == "" || strings.HasPrefix(u.Path(), "/")
+
+	// Resolve "." and ".." against a stack of real segments. Slashes (both
+	// kinds) split segments, so leading/trailing/repeated slashes normalize away.
+	var stack []string
+	splitSlashes := func(r rune) bool { return r == '/' || r == '\\' }
+	handleSegment := func(seg string) {
+		for _, seg := range strings.FieldsFunc(seg, splitSlashes) {
+			switch seg {
+			case "", ".":
+				// no-op
+			case "..":
+				if n := len(stack); n > 0 && stack[n-1] != ".." {
+					stack = stack[:n-1]
+				} else if !absolute {
+					// Can't go above the root of an absolute path; keep ".." only for relative paths.
+					stack = append(stack, "..")
+				}
+			default:
+				stack = append(stack, seg)
+			}
+		}
 	}
-	newPath += joined
-	return u.WithPath(newPath)
+	handleSegment(u.Path())
+	for _, seg := range segments {
+		handleSegment(seg)
+	}
+
+	joined := strings.Join(stack, "/")
+	if absolute {
+		joined = "/" + joined
+	}
+	return u.WithPath(joined)
 }
 
 // NewURI returns a URI from the given components.
@@ -395,7 +423,7 @@ func normalizeDriveLetter(path string) string {
 		// No drive letter found
 		return path
 	}
-	return path[0:offset] + strings.ToUpper(string(path[offset:1+offset])) + path[1+offset:]
+	return path[0:offset] + strings.ToUpper(path[offset:1+offset]) + path[1+offset:]
 }
 
 func windowsDriveLetterOffset(path string) int {
@@ -403,10 +431,16 @@ func windowsDriveLetterOffset(path string) int {
 	if strings.HasPrefix(path, "/") {
 		offset = 1
 	}
-	if len(path) >= 3+offset && path[2+offset] == '/' && path[1+offset] == ':' && ((path[0+offset] >= 'a' && path[0+offset] <= 'z') || (path[0+offset] >= 'A' && path[0+offset] <= 'Z')) {
+	if len(path) >= 3+offset && path[2+offset] == '/' && path[1+offset] == ':' && isLetter(path[0+offset]) {
 		return offset
 	}
 	return -1
+}
+
+func isLetter(c byte) bool {
+	// OR with 0x20 converts uppercase ASCII letters to lowercase
+	// Other characters will be rejected by the range check, so this is safe
+	return ((c|0x20) >= 'a' && (c|0x20) <= 'z')
 }
 
 const upperhex = "0123456789ABCDEF"
@@ -466,11 +500,6 @@ func shouldEscape(c byte, isPath, isAuthority bool) bool {
 		(isAuthority && c == ']') ||
 		(isAuthority && c == ':')
 	return !shouldKeep
-}
-
-// isLetter checks if a byte is an ASCII letter
-func isLetter(c byte) bool {
-	return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z')
 }
 
 // isDigit checks if a byte is an ASCII digit
--- a/uri.go
+++ b/uri.go
@@ -33,6 +33,10 @@ type URI interface {
 	// StringUnencoded returns the URI as a string without percent-encoding.
 	// This is useful for debugging and logging purposes.
 	StringUnencoded() string
+	// FilePath returns the URI as a file path, which is the format used by the local file system.
+	FilePath() string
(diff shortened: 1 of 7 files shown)


workspace/file_system.go
```diff
@@ -0,0 +1,328 @@
...
+var SkipDir = fs.SkipDir
+
+// SkipAll indicates that the entire file system walk should be aborted.
+var SkipAll = fs.SkipAll
+
+func WalkFileSystem(ctx context.Context, fs FileSystem, uri core.URI, fn func(DirEntry) error) error {
+	entry, err := fs.Stat(ctx, uri)
+	if err != nil {
+		return err
+	} else {
```
ssmifi: _Tiny_: Could remove the else... or keep it. :)

workspace/file_system_test.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		return nil
+	})
+	assert.Equal(t, []string{
+		"file:///a",
+		"file:///a/b",
+		"file:///a/d.txt",
+	}, visited)
+}
+
```
ssmifi, line 93: _Suggestion_: Add a `TestWalkFileSystemSkipDirOnFile` test case to test the sibling behavior.

workspace/initializer.go
```diff
@@ -35,34 +48,34 @@ func NewDefaultInitializer(sc *service.Container) Initializer {
 }
 
 func (s *DefaultInitializer) Initialize(ctx context.Context, folders []lsp.WorkspaceFolder) error {
-	if !service.Has[LanguageID](s.sc) {
+	fs, err := service.Get[FileSystem](s.sc)
+	if err != nil {
+		return nil
+	}
+	languageID, err := service.Get[LanguageID](s.sc)
```
ssmifi, line 115: _Question_: Should we continue without a `languageID`? Without `languageID`, it logs but succeeds, but in `loadFile` it then might panic, right?

workspace/initializer.go
```diff
@@ -35,34 +48,34 @@ func NewDefaultInitializer(sc *service.Container) Initializer {
...
-				return err
-			}
-			if err := ctx.Err(); err != nil {
-				return err
-			}
-			name := d.Name()
-			if d.IsDir() {
+		root := core.ParseURI(folder.URI)
+		err := WalkFileSystem(ctx, fs, root, func(entry DirEntry) error {
+			name := filepath.Base(entry.URI.Path())
```
ssmifi: _Question_: Is `filePath` OS-dependent? Does this mix well with Windows?

workspace/initializer.go
```diff
@@ -35,34 +48,34 @@ func NewDefaultInitializer(sc *service.Container) Initializer {
...
+			if entry.IsDir {
 				if strings.HasPrefix(name, ".") {
 					return filepath.SkipDir
```
ssmifi: _Question_: Should we use our local constant for consistency?

workspace/initializer.go
```diff
@@ -72,31 +85,17 @@ func (s *DefaultInitializer) Initialize(ctx context.Context, folders []lsp.Works
...
 }
 
-func (s *DefaultInitializer) loadFile(path string) {
-	content, err := os.ReadFile(path)
+func (s *DefaultInitializer) loadFile(ctx context.Context, uri core.URI, fs FileSystem, languageID LanguageID) {
+	content, err := fs.ReadFile(ctx, uri)
 	if err != nil {
-		log.Printf("failed to read file %s: %v", path, err)
+		log.Printf("failed to read file %s: %v", uri.String(), err)
 		return
```
ssmifi, line 154: _Question_: Should we report back if loading failed?

workspace/initializer.go
```diff
@@ -35,34 +48,34 @@ func NewDefaultInitializer(sc *service.Container) Initializer {
...
-			if err := ctx.Err(); err != nil {
-				return err
-			}
-			name := d.Name()
-			if d.IsDir() {
+		root := core.ParseURI(folder.URI)
+		err := WalkFileSystem(ctx, fs, root, func(entry DirEntry) error {
+			name := filepath.Base(entry.URI.Path())
+			if entry.IsDir {
 				if strings.HasPrefix(name, ".") {
```
ssmifi: This skips all hidden directories (on linux).

uri.go
```diff
@@ -210,6 +214,47 @@ func (u *uri) DocumentURI() lsp.DocumentURI {
...
+	return sb.String()
+}
+
+func (u *uri) JoinPath(segments ...string) URI {
+	if len(segments) == 0 {
+		return u
+	}
+	joined := strings.Join(segments, "/")
+	newPath := u.Path()
+	if !strings.HasSuffix(u.Path(), "/") {
```
ssmifi: _Minor_: Could use `newPath` here.

uri.go
```diff
@@ -210,6 +214,47 @@ func (u *uri) DocumentURI() lsp.DocumentURI {
...
+	}
+	if isWindows {
+		// Normalize Windows paths by replacing forward slashes with backslashes
+		path = strings.ReplaceAll(path, "/", "\\")
+	}
+	sb.WriteString(path)
+	return sb.String()
+}
+
+func (u *uri) JoinPath(segments ...string) URI {
```
ssmifi, line 246: Maybe out-of-scope for now, but `JoinPath` does not include normalization, eg, `../foo` would just be appended. Also absolute paths result in a double slash. It's fine currently, but seeing these kind of issues in other projects, it would be nice to get a solution early on. :)

uri.go
```diff
@@ -210,6 +214,47 @@ func (u *uri) DocumentURI() lsp.DocumentURI {
...
+	}
+	sb := strings.Builder{}
+	path := u.Path()
+	withAuthority := u.Authority() != ""
+	isWindows := withAuthority || windowsDriveLetterOffset(path) != -1
+	if withAuthority {
+		// UNC path: \\server\share\file.txt
+		sb.WriteString(`\\`)
+		sb.WriteString(u.Authority())
+		isWindows = true
```
ssmifi: _Minor_: `isWindows` is already true.

uri.go
```diff
@@ -325,19 +385,28 @@ func FileURI(path string) URI {
...
 }
 
 // Uppercases the drive letter in Windows file paths to ensure consistent URIs across platforms
 func normalizeDriveLetter(path string) string {
+	offset := windowsDriveLetterOffset(path)
+	if offset == -1 {
+		// No drive letter found
+		return path
+	}
+	return path[0:offset] + strings.ToUpper(string(path[offset:1+offset])) + path[1+offset:]
```
ssmifi: _Minor_: The `string` substring is superfluous.

uri.go
```diff
@@ -325,19 +385,28 @@ func FileURI(path string) URI {
...
+}
+
+func windowsDriveLetterOffset(path string) int {
 	offset := 0
 	if strings.HasPrefix(path, "/") {
 		offset = 1
 	}
-	if len(path) >= 3+offset && path[2+offset] == '/' && path[1+offset] == ':' && (path[0+offset] >= 'a' && path[0+offset] <= 'z') {
-		return path[0:offset] + strings.ToUpper(string(path[offset:1+offset])) + path[1+offset:]
+	if len(path) >= 3+offset && path[2+offset] == '/' && path[1+offset] == ':' && ((path[0+offset] >= 'a' && path[0+offset] <= 'z') || (path[0+offset] >= 'A' && path[0+offset] <= 'Z')) {
```
ssmifi: _Minor_: A little bit hard to read... it's almost 190 chars... is there no formatting in place? You could use `isLetter`, but I understand if you prefer not to. :)

uri_test.go
```diff
@@ -628,3 +628,100 @@ func TestParseUri_LowerCaseSchemeAndAuthority(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := tt.uri.JoinPath(tt.segments...)
+			assert.Equal(t, tt.expected, result.Path())
+		})
+	}
+}
```
ssmifi, line 888: _Questions_: What about `%` chars? Are these encoded/decoded correctly? Should/Could probably add a test case. Same for the lower case drive letter normalization, malformed UNC paths. Maybe also include a full round-trip: OS path -> URI string -> parsed -> FilePath

ssmifi (commented):
Looks good. I just have some minor questions about the initialization.

workspace/initializer.go
msujew: It is, but at least `Base` operates on both separators. I.e. it works even for `/` on Windows and `\` on Unix systems. Same with `Ext`.

workspace/initializer.go
msujew: You're right - probably doesn't make sense.

workspace/initializer.go
msujew: Good point 👍

workspace/initializer.go
msujew: In what way?

workspace/initializer.go
msujew: AFAIK it's the same logic as Langium (no one ever complained there). Adopters are free to adjust this logic.

workspace/file_system_test.go
msujew: I can't actually test this. The `ReadDir` ordering is non-deterministic in the virtual file system. Meaning that `SkipDir` might randomly include siblings - or not.

ssmifi: Should we sort and make it deterministic? Sure, it adds a small performance cost, but actually matches the `DiskFileSystem` behavior more closely.

workspace/initializer.go
ssmifi: I meant return the error, so that caller can react, but I just realized that `Initialize` always just returns `nil`, so it doesn't make too much sense currently.

workspace/initializer.go
ssmifi: I still don't see why we should hardcode this (platform assumption)? If there is a rationale, we should document this properly. 
Alternatively, wdyt about adding a `DirectoryFilter` with the `.` exception as default implementation to the services. Adopters then just need to change the filter.

workspace/file_system_test.go
msujew: Yeah, makes sense. I've added a sorting and some additional tests for this 👍

workspace/initializer.go
msujew: I added a new `IncludeFilter` interface.

spoenemann (approved):
Looks good! 👍

ssmifi (approved):
👍

PR #112, opened 2026-06-22
Do not create file content copy

msujew:
Resolves an issue identified in https://github.com/TypeFox/fastbelt/discussions/111. `string([]byte)` creates a copy of the byte array (as an immutable string value) which is quite wasteful on the memory side of things.

Since we never modify the byte slice anyway (we only ever reassign the pointer), creating a pointer to the byte slice as a string is a "safe" option to prevent this from happening. The only issue is that the byte slice will not be garbage collected if an adopter holds onto a value - but we already had the same problem with the string copy as well.

The following test script can be used on this branch and main (in the `benchmark_test.go` file) to confirm the improved memory consumption behavior:

<details>

```go
const resourceCount = 2000

func TestWorkspaceCycle(t *testing.T) {
	length := 0
	contents := make([]string, resourceCount)
	for i := range contents {
		contents[i], _ = generateStatemachineContent(i)
		length += len(contents[i])
	}
	srv := CreateServices()
	docManager := service.MustGet[workspace.DocumentManager](srv)
	lock := service.MustGet[workspace.Lock](srv)
	builder := service.MustGet[workspace.Builder](srv)

	docs := make([]*fastbelt.Document, resourceCount)
	for i, content := range contents {
		uri := fmt.Sprintf("file:///workspace/statemachine_%d.statemachine", i)
		doc, err := fastbelt.NewDocumentFromString(uri, "statemachine", content)
		if err != nil {
			t.Fatal(err)
		}
		docs[i] = doc
		docManager.Set(doc)
	}

	lock.Write(context.Background(), func(ctx context.Context, downgrade func()) {
		if err := builder.Build(ctx, docs, downgrade); err != nil {
			t.Errorf("build failed: %v", err)
		}
	})

	runtime.GC()
	var mem runtime.MemStats
	runtime.ReadMemStats(&mem)
	t.Errorf("heap after GC: alloc=%.1f MiB sys=%.1f MiB for %.1f MiB of code", float64(mem.HeapAlloc)/1e6, float64(mem.HeapSys)/1e6, float64(length)/1e6)

}
```

</details>

In my tests, it roughly reduces the amount of memory consumed by 10%.

### DIFF
--- a/textdoc/file.go
+++ b/textdoc/file.go
@@ -57,7 +57,9 @@ func (f *File) Version() int32 {
 	return f.version
 }
 
-// Content returns a copy of the document content as a byte slice.
+// Content returns the byte slice that represents the document content.
+// This is not a copy. Do not modify the returned slice, as it may lead
+// to unexpected behavior.
 func (f *File) Content() []byte {
 	return f.content
 }
--- a/textdoc/file.go
+++ b/textdoc/file.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"unicode/utf16"
 	"unicode/utf8"
+	"unsafe"
 
 	"typefox.dev/lsp"
 )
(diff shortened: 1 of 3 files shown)


textdoc/file.go
spoenemann, line 63: Do we need this method? With the changes of this PR, exposing content could cause follow-up issues.

Also, the comment looks inaccurate as we're not returning a copy, but the actual slice pointer.

msujew: I would keep it. We use it quite a lot internally. But I'll adjust the comment there.

spoenemann (approved):
👍

PR #113, opened 2026-06-23
Improve LSP text event handling

msujew:
Our current LSP text event handling infrastructure has a few issues:

1. It will trigger a rebuild on file close - this is not necessary, as the change event should be triggered by the client if the file changes during the close event (for example, if it resets the state of the file to the disk state).
2. When building a file, the language server always sent all diagnostics for all files included in the build to the client. It disregarded closed files.

This change addresses these two issues by improving our general LSP infrastructure to support `ServerInitializeParticipant` services that can use callbacks to properly participate in the initialization of the language server. The new `DiagnosticsPublisher` properly takes care of the state of open/closed files and updates diagnostics for all open files included in a build.

### DIFF
--- a/server/diagnostics_publisher.go
+++ b/server/diagnostics_publisher.go
@@ -1,0 +1,85 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+package server
+
+import (
+	"context"
+	"log"
+
+	core "typefox.dev/fastbelt"
+	"typefox.dev/fastbelt/textdoc"
+	"typefox.dev/fastbelt/util/service"
+	"typefox.dev/fastbelt/workspace"
+	"typefox.dev/lsp"
+)
+
+// DiagnosticsPublisher is responsible for publishing diagnostics to the LSP client.
+// The default implementation publishes diagnostics in the following scenarios:
+// Document is opened (reuses diagnostics from build).
+// Document is closed (clears diagnostics).
+// Document is validated by the builder.
+//
+// When registered to a service container, it automatically hooks into the document lifecycle events.
+type DiagnosticsPublisher struct {
+	sc *service.Container
+}
+
+// NewDiagnosticsPublisher creates a new instance of [DiagnosticsPublisher].
+func NewDiagnosticsPublisher(sc *service.Container) *DiagnosticsPublisher {
+	return &DiagnosticsPublisher{sc: sc}
+}
+
+func (d *DiagnosticsPublisher) OnServerInitialize(_ *lsp.ParamInitialize) {
+	store := service.MustGet[textdoc.Store](d.sc)
+	syncher := service.MustGet[DocumentSyncher](d.sc)
+	syncher.OnDidOpen(func(ctx context.Context, event *TextDocumentChangeEvent) {
+		docManager, err := service.Get[workspace.DocumentManager](d.sc)
+		if err != nil {
+			return
+		}
+		uri := core.ParseURI(string(event.Document.URI()))
+		document := docManager.Get(uri)
+		if document != nil {
+			// If the document already exists, publish diagnostics immediately.
+			d.publishDocumentDiagnostics(ctx, document)
+		}
+	})
+	syncher.OnDidClose(func(ctx context.Context, event *TextDocumentChangeEvent) {
+		uri := core.ParseURI(string(event.Document.URI()))
+		handle := store.Get(uri.DocumentURI())
+		if handle != nil {
+			// Clear diagnostics for the closed document.
+			d.publishDiagnostics(ctx, handle, []lsp.Diagnostic{})
+		}
+	})
+	builder := service.MustGet[workspace.Builder](d.sc)
+	builder.AddBuildStepListener(core.DocStateValidated, func(ctx context.Context, doc *core.Document) error {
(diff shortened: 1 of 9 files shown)


server/diagnostics_publisher.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (d *DiagnosticsPublisher) OnServerInitialize(_ *lsp.ParamInitialize) {
+	store := service.MustGet[textdoc.Store](d.sc)
+	syncher := service.MustGet[DocumentSyncher](d.sc)
+	syncher.OnDidOpen(func(ctx context.Context, event *TextDocumentChangeEvent) {
+		docManager, err := service.Get[workspace.DocumentManager](d.sc)
+		if err != nil {
+			return
+		}
+		uri := core.ParseURI(string(event.Document.URI()))
```
spoenemann, line 48: Why URI → string → URI?

server/diagnostics_publisher.go
```diff
@@ -0,0 +1,85 @@
...
+func (d *DiagnosticsPublisher) publishDocumentDiagnostics(ctx context.Context, doc *core.Document) {
+	lspDiags := make([]lsp.Diagnostic, 0, len(doc.Diagnostics))
+	for _, d := range doc.Diagnostics {
+		lspDiags = append(lspDiags, toLspDiagnostic(*d))
+	}
+	d.publishDiagnostics(ctx, doc.TextDoc, lspDiags)
+}
+
+func (d *DiagnosticsPublisher) publishDiagnostics(ctx context.Context, handle textdoc.Handle, diagnostics []lsp.Diagnostic) {
+	connection := service.MustGet[*Connection](d.sc)
```
spoenemann: Use Get here to avoid panic if the connection is not set?

server/document_sync.go
```diff
@@ -64,10 +78,11 @@ func NewDefaultDocumentSyncher(sc *service.Container) DocumentSyncher {
 
 func (s *DefaultDocumentSyncher) DidOpen(ctx context.Context, params *lsp.DidOpenTextDocumentParams) {
 	textdocStore := service.MustGet[textdoc.Store](s.sc)
-	existing := textdocStore.Get(params.TextDocument.URI)
+	uri := core.ParseURI(string(params.TextDocument.URI)).DocumentURI()
```
spoenemann: This line looks very strange again. If the conversion is really necessary, can we encapsulate it?

Also, the DocumentURI method name is suboptimal. Even though it's a convention to name a method after the type it converts to, it ready quite strangely. Maybe rename to LspURI, or other ideas?

server/server.go
```diff
@@ -27,10 +37,11 @@ func NewDefaultLanguageServer(sc *service.Container) lsp.Server {
 }
 
 func (s *DefaultLanguageServer) Initialize(ctx context.Context, params *lsp.ParamInitialize) (*lsp.InitializeResult, error) {
-	if slogHandler, _ := service.Get[slog.Handler](s.sc); slogHandler != nil {
-		// Set the default logger to use the configured slog handler
-		// It will send logs to the client via the LSP connection
-		slog.SetDefault(slog.New(slogHandler))
+	// Initialize all participants first
+	for service := range s.sc.All() {
+		if initializer, ok := service.(ServerInitializeParticipant); ok {
```
spoenemann: This is a multi-inject pattern. Should it have a dedicated service.Container method?

server/server.go
```diff
@@ -516,7 +527,14 @@ func StartLanguageServer(ctx context.Context, sc *service.Container) error {
 	if err != nil {
 		return err
 	}
+	store, err := service.Get[textdoc.Store](sc)
+	if err != nil {
+		return err
+	}
 	builder.AddBuildStepListener(core.DocStateValidated, func(ctx context.Context, doc *core.Document) error {
```
spoenemann: Isn't the idea of this change to eliminate the BuildStepListener from here?

server/server.go
spoenemann, line 93: We now have ServerInitializeParticipant (running on `initialize`) and workspace.Initializer (running on `initialized`). The naming might be a bit confusing...

server/server.go
```diff
@@ -8,15 +8,25 @@ import (
...
 	"typefox.dev/lsp"
 )
 
+// ServerInitializeParticipant is an interface for services that want to
+// participate in the LSP server initialization process.
+//
+// All services implementing this interface and registered in the [service.Container]
+// instance will have their [ServerInitializeParticipant.OnServerInitialize]
+// method called during the [lsp.Server.Initialize] request.
+type ServerInitializeParticipant interface {
```
spoenemann: Stuttering: server.ServerInitializeParticipant

server/diagnostics_publisher.go
msujew: For normalization. URIs from language clients don't follow a normalized schema and differ (i.e. VSCode and LSP4E use a different schema for their URIs) - and we need to normalize them into the same one that we use internally. Otherwise our lookups into the `Store` fail.

server/document_sync.go
msujew: We had to do this dance in Langium as well, see https://github.com/eclipse-langium/langium/pull/1660. But yes, we can encapsulate this.

spoenemann (approved):
👍

server/server.go
msujew: It is - I already have an idea in mind how to adjust this. We just also need a `server.InitializedParticipant`...

But that's for a separate PR ;)

PR #114, opened 2026-06-23
VS Code extension version 0.0.2

spoenemann:
Improvements of the vscode extension publishing.


internal/vscode-extensions/package.json
```diff
@@ -11,9 +11,11 @@
     "build": "npm run build --workspaces"
   },
   "devDependencies": {
-    "@types/vscode": "^1.104.0",
     "@types/node": "24.x",
+    "@types/vscode": "^1.104.0",
```
msujew: Suggestion: If we lower our required vscode version in the extension, we should lower the version here as well.

internal/vscode-extensions/fastbelt/.vscodeignore
```diff
@@ -1,6 +1,11 @@
+../**
```
msujew: Alternative: Can we simply ignore everything and provide explicit exclusions (via `!`) instead? See [here](https://github.com/zowe/zowe-pli-language-support/blob/development/packages/vscode-extension/.vscodeignore).

msujew (commented):
Thanks!

internal/vscode-extensions/fastbelt/.vscodeignore
spoenemann: I found out that `vsce ls` incorrectly lists other files from the npm workspace, but `vsce package` doesn't 🤦‍♂️

msujew (approved):

PR #116, opened 2026-06-23
Fix completion after grammar action

msujew:
Takes a tiny part of https://github.com/TypeFox/fastbelt/pull/98 to fix the completion after executing an action. Previously, synthetic objects had the wrong type after an action. That's because they used the Rule type name, instead of the actual computed interface name - which differs after actions.

Because of a minor parser generator issue if actions appear right at the start of the rule, I had to re-generate all parsers. This makes up 99% of the changes of this PR :)

The actual test for this is in the `TestCompletion_AfterO_ActionCrossRef` test.

### DIFF
--- a/internal/atn/atn.go
+++ b/internal/atn/atn.go
@@ -17,6 +17,9 @@ import (
 // completionHintFor returns the per-field CompletionHint for a CrossRef whose
 // container is an Assignment in the given rule. Returns nil if the CrossRef is
 // not nested inside an Assignment (bare cross-references contribute no hint).
+// completionHintFor returns the per-field CompletionHint for a CrossRef whose
+// container is an Assignment in the given rule. Returns nil if the CrossRef is
+// not nested inside an Assignment (bare cross-references contribute no hint).
 func completionHintFor(rule grammar.AbstractRuleWithBody, cr grammar.CrossRef) *parser.CompletionHint {
 	if rule == nil {
 		return nil
@@ -29,11 +32,16 @@ func completionHintFor(rule grammar.AbstractRuleWithBody, cr grammar.CrossRef) *
 	if prop == nil {
 		return nil
 	}
-	propName := prop.Text()
-	if propName == "" {
+	field := prop.Ref(context.Background())
+	if field == nil {
 		return nil
 	}
-	hint := &parser.CompletionHint{Field: ruleTypeName(rule) + "." + propName}
+	fieldName := field.Name()
+	if fieldName == "" {
+		return nil
+	}
+	iface := field.Container().(grammar.Interface)
+	hint := &parser.CompletionHint{Field: iface.Name() + "." + fieldName}
 	if action := findPrecedingAction(assignment); action != nil {
 		typeName := ""
 		if t := action.Type(); t != nil {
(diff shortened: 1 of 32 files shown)


internal/atn/atn.go
```diff
@@ -14,6 +14,9 @@ import (
 	"typefox.dev/fastbelt/parser"
 )
 
+// completionHintFor returns the per-field CompletionHint for a CrossRef whose
+// container is an Assignment in the given rule. Returns nil if the CrossRef is
+// not nested inside an Assignment (bare cross-references contribute no hint).
 // completionHintFor returns the per-field CompletionHint for a CrossRef whose
```
ssmifi: _Minor_: Comment is duplicated.

internal/atn/atn.go
```diff
@@ -29,11 +32,16 @@ func completionHintFor(rule grammar.AbstractRuleWithBody, cr grammar.CrossRef) *
...
+	field := prop.Ref(context.Background())
+	if field == nil {
+		return nil
+	}
+	fieldName := field.Name()
+	if fieldName == "" {
 		return nil
 	}
-	hint := &parser.CompletionHint{Field: ruleTypeName(rule) + "." + propName}
+	iface := field.Container().(grammar.Interface)
```
ssmifi: _Question_: Is it possible for the container not being a `grammar.Interface`? I assume, in practice, any field should have an interface, but is this also safe, e.g., at malformed grammars, or should it be guarded?

internal/generator/type_generator.go
```diff
@@ -407,3 +412,36 @@ func generateDataStruct(node codegen.Node, iface grammar.Interface, fields []Fie
...
+	entries := []factoryEntry{}
+	for _, iface := range grammr.Interfaces() {
+		entries = append(entries, factoryEntry{key: iface.Name(), typeName: iface.Name()})
+	}
+	slices.SortStableFunc(entries, func(a, b factoryEntry) int {
+		return cmp.Compare(a.key, b.key)
+	})
+
+	// Sort for stable output.
+	sort.Slice(entries, func(i, j int) bool { return entries[i].key < entries[j].key })
```
ssmifi, line 431: _Question_: Aren't these sorting the same slice?

internal/generator/parser_generator.go
```diff
@@ -879,7 +879,10 @@ func generateParseFunction(node codegen.Node, context *ParserGeneratorContext, r
 		node.Indent(func(n codegen.Node) {
 			n.AppendLine("current := New", returnType.Name(), "()")
 			n.AppendLine("current.SetSegmentStartToken(p.state.LA(1))")
+			// Generate new lexical scope for actions that immediately trigger on tule start
```
ssmifi: _Typo_: "tule"

ssmifi (approved):
LGTM, just a guarding question and some minor remarks.

internal/generator/type_generator.go
msujew: They are - I shouldn't blindly copy other people's code 😆 (see #98):

https://github.com/TypeFox/fastbelt/blob/10ec5119675f5f774159236453d0a4a426528251/internal/generator/type_generator.go#L444-L450

internal/generator/parser_generator.go
msujew: That's how you can tell I wrote this code myself 🙃

internal/atn/atn.go
msujew: Not really possible - in fact, none of these guards should ever be possible to hit in a validated grammar. I've changed them all to `panic`, since they represent invalid grammar states.

PR #117, opened 2026-06-24
Properly expect EOF after parser exits

msujew:
While we are perfectly capable of telling whether an input is too short (because parsing it fails), we're currently too lenient when an input is "too long" (i.e. the parser has already exited, but there are more tokens available).

This change addresses this by adding a dedicated `ExpectEndOfInput` method to the parser state, which generates an error on the first extra token it sees after the parser exits.

### DIFF
--- a/parser/error_recovery.go
+++ b/parser/error_recovery.go
@@ -41,13 +41,13 @@ func NewBailErrorRecovery() BailErrorRecovery {
 
 func (DefaultErrorRecovery) RecoverInline(parserState *ParserState, expectedTokenType *core.TokenType) (*core.Token, bool) {
 	la1 := parserState.LA(1)
-	if la1 == nil {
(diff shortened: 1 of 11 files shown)


parser/error_recovery.go
Lotes, line 57: Update the comment? You are refering to `nil` tokens. I am not sure if this is still given, since you changed the if conditions.

Lotes (approved):
Looks straight forward. Thanks.

parser/error_recovery.go
msujew: `RecoverInline` is called in `ParserState.Consume`. Consume still returns `nil` tokens, when they don't match the expected token - the comment is therefore still correct. Note that `RecoverInline` actually returns `nil`, not the EOF token.

PR #120, opened 2026-06-24
Grammar improvements: validations, optional semicolons, entry rules

### LINKED ISSUE #119: [Grammar] Implement additional grammar validations
There are a few grammar validations currently missing. Most notably:

* [x] Some field names (such as `Tokens` and `Document`) are reserved by the framework. There's no validation that guards adopters - which will then simply result in a Go compiler error.
* [x] Hidden tokens and comment tokens cannot be used (reasonably) for cross references - the grammar currently allows that.
* [x] Arrays for interface fields can be declared as nested - this doesn't make sense, since the parser cannot generate nested array fields (i.e. `Field [][]string`)
* [x] Composite rules can be incorrectly assigned to elements of type `string` (they should use `composite` instead).
* [x] Our code generator always expects a cross-reference terminal. Our parser doesn't right now. Meaning that a cross reference like `Ref=[Some]` passes the validation, but panics during code generation.

spoenemann:
- Adds missing validations and fixes #119.
- Makes semicolons optional. ~_Note:_ for parser rules and composite rules, a semicolon is still needed if another parser rule follows after that because it starts with an ID. Maybe we can improve the lookahead for this specific case later.~
- Consistently call grammar interface fields "fields" instead of "properties" in error messages.
- _Added:_ New `entry` keyword to explicitly mark the grammar's entry rule

### DIFF
--- a/internal/grammar/validator.go
+++ b/internal/grammar/validator.go
@@ -16,24 +16,40 @@ import (
 )
 
 const (
-	ValidateUniqueRuleName         = "uniqueRuleName"
-	ValidateUniqueInterfaceName    = "uniqueInterfaceName"
-	ValidateEmptyToken             = "emptyTerminalRule"
-	ValidateEmptyKeyword           = "emptyKeyword"
-	ValidateWhitespaceOnlyKeyword  = "whitespaceOnlyKeyword"
-	ValidateKeywordWithWhitespace  = "keywordWithWhitespace"
-	ValidateRuleReturnType         = "ruleReturnType"
-	ValidateInterfaceExtends       = "interfaceExtends"
-	ValidateRuleCallReturnType     = "ruleCallReturnType"
-	ValidateRuleCallPosition       = "ruleCallPosition"
-	ValidateActionAssignmentType   = "actionAssignmentType"
-	ValidateActionPropertyType     = "actionPropertyType"
-	ValidateAssignmentType         = "assignmentType"
-	ValidateRecursiveTokenGroup    = "recursiveTokenGroup"
-	ValidateInvalidTokenInGroup    = "invalidTokenInGroup"
-	ValidateUniqueFieldName        = "uniqueFieldName"
-	ValidateFieldNameCapitalLetter = "fieldNameCapitalLetter"
+	ValidateUniqueRuleName          = "uniqueRuleName"
+	ValidateUniqueInterfaceName     = "uniqueInterfaceName"
+	ValidateEmptyToken              = "emptyTerminalRule"
+	ValidateEmptyKeyword            = "emptyKeyword"
+	ValidateWhitespaceOnlyKeyword   = "whitespaceOnlyKeyword"
+	ValidateKeywordWithWhitespace   = "keywordWithWhitespace"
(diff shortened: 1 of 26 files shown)


msujew:
> Makes semicolons optional. Note: for parser rules and composite rules, a semicolon is still needed if another parser rule follows after that because it starts with an ID. Maybe we can improve the lookahead for this specific case later.

You can change the decision for that to use LL instead of SLL. That way, it will parse correctly (at the cost of a bit of performance).

spoenemann:
>> Makes semicolons optional. Note: for parser rules and composite rules, a semicolon is still needed if another parser rule follows after that because it starts with an ID. Maybe we can improve the lookahead for this specific case later.
>
> You can change the decision for that to use LL instead of SLL. That way, it will parse correctly (at the cost of a bit of performance).

Ok like [this](https://github.com/TypeFox/fastbelt/pull/120/changes#diff-24d809afbf7a62933208baa02e573d71232c74309cb787d791ca11f0aae93efd)?

msujew:
> Ok like [this](https://github.com/TypeFox/fastbelt/pull/120/changes#diff-24d809afbf7a62933208baa02e573d71232c74309cb787d791ca11f0aae93efd)?

I originally had something different in mind - but that doesn't seem to work correctly (for some reason). Your solution works well 👍

internal/grammar/validator.go
```diff
@@ -16,25 +16,41 @@ import (
...
+	ValidateActionPropertyType      = "actionPropertyType"
+	ValidateAssignmentType          = "assignmentType"
+	ValidateRecursiveTokenGroup     = "recursiveTokenGroup"
+	ValidateInvalidTokenInGroup     = "invalidTokenInGroup"
+	ValidateInvalidTokenInCrossRef  = "invalidTokenInCrossRef"
+	ValidateMissingCrossRefTerminal = "missingCrossRefTerminal"
+	ValidateUniqueFieldName         = "uniqueFieldName"
+	ValidateFieldNameCapitalLetter  = "fieldNameCapitalLetter"
+	ValidateReservedFieldName       = "reservedFieldName"
+	ValidateNestedArrayType         = "nestedArrayType"
```
msujew, line 39: Suggestion: While you're at it, I noticed that the language ID isn't used as the `Source` property of the diagnostics yet. Can you adjust the `DefaultDocumentValidator` to add that as the source, if none is set?

msujew (approved):
Thanks, that helps a lot. Just one comment - also I want some feedback for the optional `;` at the end of rules. @cdietrich care to chime in? WDYT?

workspace/document_validator.go
```diff
@@ -56,5 +56,26 @@ func (s *DefaultDocumentValidator) Validate(ctx context.Context, doc *core.Docum
 			validator.Validate(ctx, level, accept)
 		}
 	}
+	setDiagnosticSource(s.sc, doc, diagnostics)
```
msujew: Suggestion: Move into `accept` function and precompute the `LanguageID` in this method. This way we don't need to re-iterate all diagnostics.

PR #122, opened 2026-06-25
Update version of Fastbelt extension to 0.1.0

msujew:
In preparation for the 0.1 release. I'll create a tag on the merge commit afterwards.


spoenemann (approved):
Thanks! Anything else we should think about for the release? 🤔

msujew:
> Anything else we should think about for the release? 🤔

I don't think so.

PR #124, opened 2026-06-30
AstNode locating

sailingKieler:
Contributes
* AstNode path fragment computation and
* AstNode identification based on a given path and a parent/root node.

Most significant changes:
* addition of fields `containerField` and `containerIndex` to `AstNodeBase`
  * `containerField` takes `unique.Handle[string]` representing pointers of interned string constants representing the field names
  * `containerIndex` takes an `int`
* addition of func `ContainmentData()` to `AstNode` and `AstNodeBase` providing the former fields' values
* extension of `SetContainer()` to also expect arguments for the former fields and assigns those ones
* extension of `ForEachNode()` and `ForEachReference()`, precisely extension of the expected lambdas providing the former values, too; required for recursive application of `SetContainer(...)` within `AssignContainers(...)`
* addition of (static) funcs `PathOf(AstNode) (string, error)` and `Resolve(AstNode, string) (AstNode, error)`
  * computes the path fragment of an AstNode within its parents,
* addition of path string abstraction interface named `FragmentPath` + implementation `fragmentPath`
* addition of func `Resolve(...)` to `AstNode` and `AstNodeBase` (base impl) + generation for each impl type
  * reveals a nested ast node within some parent node according to the given (relative) path
* generation of interned string handles for each occurring field name
* generation of type specific impls of `Resolve(FragmentPath)`
* error handling/reporting
* unit tests in `ast_locating_test.go` within the `Arithmetics` example language
* benchmarks in `ast_locating_benchmark_test.go` within the `Arithmetics` example language

### DIFF
--- a/ast.go
+++ b/ast.go
@@ -27,14 +27,14 @@ type AstNode interface {
 	// Container returns the direct parent node of the node in the AST.
 	// It returns nil if this is the root node.
 	Container() AstNode
+	// ContainmentData returns a [unique.Handle] denoting the containing property within it's [AstNode.Container],
+	// defaults to a a [unique.Handle] of the empty string,
+	// and the element index within the containing property, defaults to zero for single item fields
+	ContainmentData() (unique.Handle[string], uint16)
 	// SetContainer sets the direct parent node of the node.
 	//
 	// When constructing an AST programmatically, use [AssignContainers] to link the node in the AST.
 	SetContainer(container AstNode, containerField unique.Handle[string], index uint16)
-	// NodePath returns a slash-separated path string that uniquely identifies this node
-	// within its document tree, e.g. "rules@2/alternatives@0".
-	// Returns "" for the root node (no container).
-	NodePath() (string, error)
 	// Tokens returns the tokens associated with the node.
 	Tokens() []*Token
 	// SetToken appends token to the node's token list.
@@ -67,6 +67,18 @@ type AstNode interface {
 	//
 	// Calling this method directly is not recommended. Use [References] instead for better readability.
 	ForEachReference(fn func(UntypedReference, unique.Handle[string], uint16))
+	// FieldInfos returns a descriptor for the denoted field of this AstNode.
+	FieldInfos(field unique.Handle[string]) FieldInfos
+	// NodePath returns a slash-separated path string that uniquely identifies this node
+	// within its document tree, e.g. "rules@2/alternatives@0".
+	// Returns "" for the root node (no container).
+	NodePath() (string, error)
+}
+
+// FieldInfos is a simple struct of meta data describing a field of an AstNode.
+type FieldInfos struct {
+	Multi     bool
+	Reference bool
 }
 
 // AstNodeBase provides the default [AstNode] implementation used by generated AST node types.
@@ -95,63 +107,6 @@ func (node *AstNodeBase) SetDocument(document *Document) {
 	}
 }
 
-type FieldInfos struct {
-	Multi     bool
-	Reference bool
-}
-
-// Base Implementation returning field meta data for an AstNode.
-// The generator produces specific overwrites for each generated ...Impl type.
-func (node *AstNodeBase) FieldInfos(field unique.Handle[string]) FieldInfos {
-	return FieldInfos{
-		Multi:     false,
-		Reference: false,
-	}
-}
-
-func (node *AstNodeBase) NodePath() (string, error) {
-	return GetNodePath(node)
-}
-
-type withFieldInfos interface {
-	ContainerData() (unique.Handle[string], uint16)
-	FieldInfos(field unique.Handle[string]) FieldInfos
-}
-
-func GetNodePath(node AstNode) (string, error) {
-	impl, ok := node.(withFieldInfos)
-	if !ok {
-		return "", errors.New("GetNodePath: conversion failed")
-	}
-
-	container := node.Container()
-	containerField, index := impl.ContainerData()
-
-	if container == nil {
-		return "", nil
-	} else if containerField.Value() == "" {
-		return "", errors.New("Can't determine node path, found AstNode field 'containerField' being empty.")
-	}
-
-	parentPath, err := GetNodePath(container)
-
-	if err == nil {
-		fieldPath := parentPath + "/" + containerField.Value()
-		containerWithInfos, ok := container.(withFieldInfos)
-		if ok && containerWithInfos.FieldInfos(containerField).Multi {
-			return fieldPath + "@" + strconv.Itoa(int(index)), nil
-		} else if ok {
-			x := containerWithInfos.FieldInfos(containerField)
-			println(x.Multi)
-			return fieldPath, nil
-		} else {
-			return "", errors.New("GetNodePath: conversion failed")
-		}
-	} else {
-		return "", err
-	}
-}
-
 // Container returns the direct parent node of the node in the AST.
 // It returns nil if this is the root node.
 func (node *AstNodeBase) Container() AstNode {
@@ -162,7 +117,7 @@ func (node *AstNodeBase) Container() AstNode {
 	}
 }
 
-func (node *AstNodeBase) ContainerData() (unique.Handle[string], uint16) {
+func (node *AstNodeBase) ContainmentData() (unique.Handle[string], uint16) {
 	return node.containerField, node.containerIndex
 }
 
@@ -259,6 +214,37 @@ func (node *AstNodeBase) Text() string {
 	} else {
 		fullText := node.document.TextDoc.Text(nil)
 		return fullText[node.segment.Indices.Start:node.segment.Indices.End]
+	}
+}
+
+// Base Implementation returning field meta data for an AstNode.
+// The generator produces specific overwrites for each generated ...Impl type.
+func (node *AstNodeBase) FieldInfos(field unique.Handle[string]) FieldInfos {
+	return FieldInfos{}
+}
+
+// NodePath determines node's path with it's root container in a recursive manner,
+// based on node's [AstNodeBase.ContainmentData] and [AstNode.FieldInfos]
+func (node *AstNodeBase) NodePath() (string, error) {
+	container := node.Container()
+	containerField, index := node.ContainmentData()
+
+	if container == nil {
+		return "", nil
+	} else if containerField.Value() == "" {
+		return "", errors.New("Can't determine node path, found AstNode's field 'containerField' being empty.")
+	}
+
+	parentPath, err := container.NodePath()
+	if err != nil {
+		return "", err
+	}
+
+	fieldPath := parentPath + "/" + containerField.Value()
+	if container.FieldInfos(containerField).Multi {
+		return fieldPath + "@" + strconv.Itoa(int(index)), nil
+	} else {
+		return fieldPath, nil
 	}
 }
 
--- a/ast.go
+++ b/ast.go
@@ -73,6 +73,8 @@ type AstNode interface {
 	// within its document tree, e.g. "rules@2/alternatives@0".
 	// Returns "" for the root node (no container).
 	NodePath() (string, error)
+	// GetByPath returns a (nested) child node denoted by the given path
+	GetByPath(path string) (AstNode, error)
 }
 
 // FieldInfos is a simple struct of meta data describing a field of an AstNode.
@@ -245,6 +247,16 @@ func (node *AstNodeBase) NodePath() (string, error) {
 		return fieldPath + "@" + strconv.Itoa(int(index)), nil
 	} else {
 		return fieldPath, nil
+	}
+}
+
+// Base Implementation for instances of [AstNodeBase].
+// The generator produces specific overwrites for each generated ...Impl type.
+func (node *AstNodeBase) GetByPath(path string) (AstNode, error) {
+	if path != "" {
+		return nil, errors.New("AstNodeBase.GetByPath: Cannot identify children of plain AstNodeBase instances.")
+	} else {
+		return node, nil
 	}
 }
 
--- a/ast.go
+++ b/ast.go
@@ -234,7 +234,7 @@ func (node *AstNodeBase) NodePath() (string, error) {
 	if container == nil {
 		return "", nil
 	} else if containerField.Value() == "" {
-		return "", errors.New("Can't determine node path, found AstNode's field 'containerField' being empty.")
+		return "", errors.New("cannot determine node path, found AstNode's field 'containerField' being empty")
 	}
 
 	parentPath, err := container.NodePath()
@@ -254,7 +254,7 @@ func (node *AstNodeBase) NodePath() (string, error) {
 // The generator produces specific overwrites for each generated ...Impl type.
 func (node *AstNodeBase) GetByPath(path string) (AstNode, error) {
 	if path != "" {
-		return nil, errors.New("AstNodeBase.GetByPath: Cannot identify children of plain AstNodeBase instances.")
+		return nil, errors.New("AstNodeBase.GetByPath: Cannot identify children of plain AstNodeBase instances")
 	} else {
 		return node, nil
 	}
--- a/ast.go
+++ b/ast.go
@@ -6,6 +6,7 @@ package fastbelt
 
 import (
 	"errors"
+	"fmt"
 	"iter"
 	"strconv"
 	"strings"
@@ -225,6 +226,8 @@ func (node *AstNodeBase) FieldInfos(field unique.Handle[string]) FieldInfos {
 	return FieldInfos{}
 }
 
+var fieldZero = unique.Handle[string]{}
+
 // NodePath determines node's path with it's root container in a recursive manner,
 // based on node's [AstNodeBase.ContainmentData] and [AstNode.FieldInfos]
 func (node *AstNodeBase) NodePath() (string, error) {
@@ -233,13 +236,19 @@ func (node *AstNodeBase) NodePath() (string, error) {
 
 	if container == nil {
 		return "", nil
-	} else if containerField.Value() == "" {
-		return "", errors.New("cannot determine node path, found AstNode's field 'containerField' being empty")
+	} else if containerField == fieldZero || containerField.Value() == "" {
+		return "", errors.New("cannot determine node path, 'containerField' is empty")
 	}
 
 	parentPath, err := container.NodePath()
 	if err != nil {
-		return "", err
+		if errors.Unwrap(err) == nil {
+			return "", fmt.Errorf(
+				"AstNodeBase.NodePath: error within node of type %T: %w", container, err)
+		} else {
+			return "", fmt.Errorf(
+				"AstNodeBase.NodePath: error within container of type %T:\n %w", container, err)
+		}
 	}
 
 	fieldPath := parentPath + "/" + containerField.Value()
--- a/ast.go
+++ b/ast.go
@@ -5,9 +5,12 @@
 package fastbelt
 
 import (
+	"errors"
 	"iter"
+	"strconv"
 	"strings"
 	"sync/atomic"
+	"unique"
 )
 
 // AstNode is the base interface for all AST nodes.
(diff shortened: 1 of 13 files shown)


ast.go
```diff
@@ -55,19 +63,35 @@ type AstNode interface {
...
+	// Returns "" for the root node (no container).
+	NodePath() (string, error)
+	// GetByPath returns a (nested) child node denoted by the given path
+	GetByPath(path string) (AstNode, error)
+}
+
+// FieldInfos is a simple struct of meta data describing a field of an AstNode.
+type FieldInfos struct {
+	Multi     bool
+	Reference bool
```
msujew: Question: What is the `Reference` field for?

ast.go
```diff
@@ -55,19 +63,35 @@ type AstNode interface {
...
 // AstNodeBase provides the default [AstNode] implementation used by generated AST node types.
 type AstNodeBase struct {
-	document  *Document
-	container AstNode
-	tokens    []*Token
-	segment   TextSegment
+	document       *Document
+	container      AstNode
+	containerField unique.Handle[string]
+	containerIndex uint16
```
msujew: Suggestion: You can make this `containerIndex` an `int` (which is 4 or 8 bytes, depending on the target architecture). No difference in Go struct size. Go optimizes layouts for CPU stride and since everything is now 64 bits, most structs are layouted in 8 byte intervals (there are exceptions, see `TextRange` for example). Note that I specifically recommend using `int` here to also set negative values. See the other recommendation below.

ast.go
```diff
@@ -55,19 +63,35 @@ type AstNode interface {
...
 	// Calling this method directly is not recommended. Use [ChildNodes] instead for better readability.
-	ForEachNode(fn func(AstNode))
+	ForEachNode(fn func(AstNode, unique.Handle[string], uint16))
 	// ForEachReference calls fn for each reference field of node.
 	//
 	// Calling this method directly is not recommended. Use [References] instead for better readability.
-	ForEachReference(fn func(UntypedReference))
+	ForEachReference(fn func(UntypedReference, unique.Handle[string], uint16))
+	// FieldInfos returns a descriptor for the denoted field of this AstNode.
+	FieldInfos(field unique.Handle[string]) FieldInfos
```
msujew: Suggestion: I don't think we actually need this. As far as I could tell, we only require this information to determine whether a property was part of an array to generate the `@<index>` suffix. However, if we can make this clear in the `containerIndex` already, we need none of this information (especially because `FieldInfos.Reference` seems to be unused. We could either:
1. Say that `containerIndex == -1` indicates that the AstNode is not part of an array
2. Say that `containerIndex == 0` indicates that, but then every `containerIndex` needs to be offset by `+1`.

ast.go
```diff
@@ -55,19 +63,35 @@ type AstNode interface {
...
+	// NodePath returns a slash-separated path string that uniquely identifies this node
+	// within its document tree, e.g. "rules@2/alternatives@0".
+	// Returns "" for the root node (no container).
+	NodePath() (string, error)
```
msujew: Discussion: I'm wondering whether it makes sense to introduce an abstraction of the `NodePath`. I.e. a `Path` struct with:

```go
func ParsePath(string) Path { ... }

type Path struct {
  Segments []PathSegment
}

func (p Path) String() string {...}

type PathSegment struct {
  Field string
  Index int
}
```

As part of this, I would actually rename this property to `Path`. That would alleviate the need to do that much adhoc string parsing inside of the generated `GetByPath` code.

ast.go
```diff
@@ -55,19 +63,35 @@ type AstNode interface {
...
+	// GetByPath returns a (nested) child node denoted by the given path
+	GetByPath(path string) (AstNode, error)
```
msujew: Suggestion: Go usually doesn't use getters in interfaces. Maybe `ByPath(path Path)` or `FromPath`?

ast.go
```diff
@@ -190,17 +220,66 @@ func (node *AstNodeBase) Text() string {
...
+		} else {
+			return "", fmt.Errorf(
+				"AstNodeBase.NodePath: error within container of type %T:\n %w", container, err)
+		}
+	}
+
+	fieldPath := parentPath + "/" + containerField.Value()
+	if container.FieldInfos(containerField).Multi {
+		return fieldPath + "@" + strconv.Itoa(int(index)), nil
+	} else {
```
msujew: Suggestion: This amount of string concatenation is rather expensive. Can we replace this with a collection phase + `strings.Builder`?

msujew (commented):
Note that this introduces new methods to the `AstNode`, which are now reserved. You'll need to adjust the grammar validator, to prevent compilation errors down the line.

ast.go
sailingKieler: Good idea. Introduced interface `FragmentPath` in `path.go`.

ast.go
sailingKieler: IMO your proposals are in conflict with the goal of not blocking potential language type field names,
which should have higher priority I think.
Anyway, I renamed the function to 'Resolve', which is hopefully _imperative_ enough to not be a common field name candidate.

ast.go
sailingKieler: done.

ast.go
sailingKieler: Chose version 1.

ast.go
sailingKieler: done.

ast.go
sailingKieler: removed the entire thing.

sailingKieler:
Now ready for another review, @msujew.

path.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+type FragmentPath interface {
+	Empty() bool
+	Push(name unique.Handle[string], index int) FragmentPath
+	Shift() (unique.Handle[string], int)
+	String() string
+}
+
+type FragmentPathImpl struct {
```
spoenemann: Proposal: make this private; if adopters might want to instantiate it directly, add a `New...` constructor function.

path.go
```diff
@@ -0,0 +1,124 @@
...
+// NodePath determines node's path within its root container in a recursive manner,
+// based on node's [AstNode.ContainmentData]. It returns a slash-separated
+// path string that uniquely identifies this node within its document tree,
+// e.g. "/rules@2/alternatives@0".
+// Returns "" for the root node (no container).
+func PathOf(node AstNode) (FragmentPath, error) {
+	container := node.Container()
+	containerField, index := node.ContainmentData()
+
+	b := &FragmentPathImpl{}
```
spoenemann: `b` is used only in case of early exit – it should be kept out of the happy path to avoid unnecessary instantiation.

examples/statemachine/types_gen.go
```diff
@@ -123,14 +126,69 @@ type StatemachineImpl struct {
...
+func (i *StatemachineImpl) ForEachNode(fn func(core.AstNode, unique.Handle[string], int)) {
 	i.StatemachineData.ForEachNode(fn)
 }
 
-func (i *StatemachineImpl) ForEachReference(fn func(core.UntypedReference)) {
+func (i *StatemachineImpl) ForEachReference(fn func(core.UntypedReference, unique.Handle[string], int)) {
 	i.StatemachineData.ForEachReference(fn)
 }
 
+func (i *StatemachineImpl) Resolve(path core.FragmentPath) (core.AstNode, error) {
```
spoenemann, line 137: Calling `Resolve` with an empty path should return itself (`i`). This means that the `if path.Empty()` check before recursion can be moved to the beginning of the method.

examples/statemachine/types_gen.go
```diff
@@ -123,14 +126,69 @@ type StatemachineImpl struct {
...
+	field, index := path.Shift()
```
spoenemann: This is a destructive operation, which can be a potential source of error: if I call `Resolve`, the path that I pass as argument becomes unusable afterwards. I propose splitting this into separate non-destructive `Head` and `Tail` operations.

path.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"unique"
+)
+
+type FragmentPath interface {
```
spoenemann, line 16: Please add comments to all public elements following the Go doc standards. If you'd like to generate the comments, I recommend using [go-documentation](https://github.com/TypeFox/agent-skills/tree/main/skills/go-documentation) as supporting skill.

Method implementations don't need documentation except if there's something worth mentioning about how it's implemented.

examples/statemachine/types_gen.go
sailingKieler: The idea is to use the destructive operation while navigating along the path.
The path object is supposed to be temporary, the actual API entry point is `[core.]Resolve(string, AstNode)` in `path.go`.
Instead of changing the current path object the function could create a copy.
However, is that worth the effort?

path.go
sailingKieler: I'm about it already.

examples/statemachine/types_gen.go
sailingKieler: Also an option, yes.

examples/statemachine/types_gen.go
spoenemann: The copy is not more than a pointer to a slice and the number of copies equals the tree depth – I assume this is negligible, but we can benchmark it if in doubt.

This is not a hot code path (not used anywhere by the framework yet), so I suggest to prioritize a clean API.

path.go
spoenemann: About the name: I believe "Fragment" originates in the term "URI fragment" which was used a lot in EMF. In itself, the concept is independent of URIs, so it could make sense to give it a different name that clarifies what kind of path it is, e.g. `NodePath` or `AstNodePath`.

path.go
sailingKieler: My idea went into the direction of giving adaptors the chance to re-use and modify it.
However, it's so simple -- copying won't hurt much if really needed.

Changed it to non-exported.

path.go
sailingKieler: It was a mixed of good cases (no container) and erroneous cases -- and partly a left over from earlier drafts.

Refined it.

examples/statemachine/types_gen.go
sailingKieler: Had to think about for a moment, it doesn't look bad. Did it that way.

path.go
sailingKieler: We have it here, too: https://github.com/TypeFox/fastbelt/blob/9d95c2126f009e30d32ff13a39de85b8422deb21/uri.go#L52-L53

Therefore I used the term here, too.

sailingKieler:
Looking forward to more suggestions.

path.go
```diff
@@ -0,0 +1,145 @@
...
+	// String returns the corresponding fragment path string
+	String() string
+}
+
+// Base implementation of [FragmentPath]
+type fragmentPathImpl struct {
+	segments []struct {
+		name  unique.Handle[string]
+		index int
+	}
```
msujew: Suggestion: I would create a dedicated type for this. It is rather unusual in Go to have inline type definitions in other types. It's usually used for types that are local to a specific function.

path.go
```diff
@@ -0,0 +1,145 @@
...
+// Returns an empty descriptor for root nodes/nodes without a configured container,
+// its [FragmentPath.String]() yields an empty string.
+func PathOf(node AstNode) (*fragmentPathImpl, error) {
+	container := node.Container()
+	containerField, index := node.ContainmentData()
+
+	if container == nil {
+		return &fragmentPathImpl{}, nil
+	} else if containerField == fieldZero || containerField.Value() == "" {
+		return nil, errors.New("cannot determine node path, 'containerField' is empty")
```
msujew, line 88: Minor: Is the `Value()` call here actually necessary? Or does it make more sense to store an `fieldEmpty = unique.Make("")` that we can use to compare?

path.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (p *fragmentPathImpl) String() string {
+	var sb strings.Builder
+	for _, s := range p.segments {
+		sb.WriteString("/")
+		sb.WriteString(s.name.Value())
+		if s.index >= 0 {
+			sb.WriteString("@")
+			sb.WriteString(strconv.Itoa(int(s.index)))
```
msujew: Minor: Isn't that already a `int`?

```suggestion
			sb.WriteString(strconv.Itoa(s.index))
```

path.go
```diff
@@ -0,0 +1,145 @@
...
+			return nil, fmt.Errorf(
+				"PathOf: error within container of type %T:\n %w", container, err)
+		}
+	}
+	return parentPath.push(containerField, index), nil
+}
+
+// Resolve returns the (deeply) contained child of node denoted by path.
+// Leading slashes of path are ignored, the field names are evaluated child by child starting with node.
+func Resolve(path string, node AstNode) (AstNode, error) {
```
msujew: Minor: I feel like the order of parameters should be reversed here:
```suggestion
func Resolve(node AstNode, path string) (AstNode, error) {
```

internal/generator/type_generator.go
```diff
@@ -440,3 +447,144 @@ func generateSyntheticFactories(grammr grammar.Grammar) codegen.Node {
...
+	seen := map[string]bool{}
+	var pnames []string
+	for _, iface := range grammr.Interfaces() {
+		for _, f := range collectAllFields(iface, map[string]struct{}{}) {
+			if !seen[f.PName] {
+				seen[f.PName] = true
+				pnames = append(pnames, f.PName)
+			}
+		}
+	}
```
msujew: Minor: By now, we have a `Set` interface for this, I believe.

examples/arithmetics/ast_locating_test.go
msujew, line 1: Suggestion: Can you add benchmarks for generation of paths + another one for retrieval via path here?

ast.go
```diff
@@ -24,10 +26,14 @@ type AstNode interface {
 	// Container returns the direct parent node of the node in the AST.
 	// It returns nil if this is the root node.
 	Container() AstNode
+	// ContainmentData returns a [unique.Handle] denoting the containing property within it's [AstNode.Container],
+	// defaults to a a [unique.Handle] of the empty string,
+	// and the element index within the containing property, defaults to zero for single item fields
```
msujew:
```suggestion
	// ContainmentData returns a [unique.Handle] denoting the containing property within it's [AstNode.Container],
	// defaults to a [unique.Handle] of the empty string,
	// and the element index within the containing property, defaults to -1 for single item fields
```

path.go
```diff
@@ -0,0 +1,145 @@
...
+
+// Base implementation of [FragmentPath]
+type fragmentPathImpl struct {
+	segments []struct {
+		name  unique.Handle[string]
+		index int
+	}
+}
+
+var _ FragmentPath = (*fragmentPathImpl)(nil)
```
msujew: Question: What is this for?

path.go
msujew: Question: Can this be declared as:

```suggestion
type segment struct {
    name  unique.Handle[string]
    index int
}
// Base implementation of [FragmentPath]
type fragmentPath segment[]
```

And then have all methods operate on `fragmentPath`?

msujew (commented):
Thanks, the API looks really nice now!

path.go
sailingKieler: That's a so called _compile-time satisfaction check_, Claude came up with this:
> It's a canary. Without it, the compiler only complains about the interface mismatch at the point where you assign or pass a *FragmentPathImpl as a FragmentPath — which may be far from the type definition, or (in a library) not even in your own package.
>
> With the blank assignment near the type definition, the compiler tells you immediately, at the declaration site, which method is missing or has the wrong signature rather than somewhere downstream.

So I gave it a try.

internal/generator/type_generator.go
sailingKieler: Does it have any advantage? This one is the AI proposed solution I think.

path.go
msujew: Ah, I see. Makes sense 👍

internal/generator/type_generator.go
msujew: It's shorter. Since you can just do:

```suggestion
	set := core.NewSet[string]()
	for _, iface := range grammr.Interfaces() {
		for _, f := range collectAllFields(iface, map[string]struct{}{}) {
			set.Add(f.PName)
		}
	}
	pnames := slices.Collect(set.All())
```

sailingKieler: But mine is _idiomatic_ 😛

path.go
sailingKieler: From the pure user point of view `fastbelt.Resolve("/statements@7/expression/left/right", root)` makes more sense to me, as you can read it like "fastbelt.Resolve child "/statements@7/expression/left/right" in parent root".

path.go
sailingKieler: I discussed this with Claude, there're pros and cons:
> __On performance__: containerField == fieldEmpty is a single pointer comparison — the whole point of unique. containerField.Value() == "" dereferences the pointer first, then does a string comparison. For an empty string the comparison is trivially O(1) (length 0), but the pointer dereference still happens. The difference is negligible in practice.
> [...]
>
> Replacing the second clause with containerField == fieldEmpty is still valid — the nil dereference is already prevented by the short-circuit — but given you're already accepting that external callers may return a zero handle, they could equally return unique.Make("") or some other handle whose .Value() is "". The .Value() == "" form is arguably more explicit about the intent: reject any handle whose string value is empty, regardless of how it was constructed.
>
> So the current code is the right call given your constraints. No change needed there.

path.go
sailingKieler: Probably. Initially also thought into that direction, but didn't want to over complicate it.
Now that you asked I didn't have an excuse anymore...

internal/generator/type_generator.go
sailingKieler: replaced.

path.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"unique"
+)
+
+// FragmentPath describes instance of node path descriptors pointing to a child AST node within a parent.
```
spoenemann: Minor: something's missing grammatically – maybe "FragmentPath is a node path descriptor..."?

path.go
```diff
@@ -0,0 +1,137 @@
...
+type FragmentPath interface {
+	// Empty returns [true] if this descriptor is empty and equal to the empty path string, returns [false] otherwise.
```
spoenemann: The `[...]` syntax is used for links to other symbols, e.g. `[PathOf]`. Code should be written without any delimiters in comments.

spoenemann (approved):
This looks really good to me, great work!

path.go
msujew: Interesting. I'd read `fastbelt.Resolve(root, "/statements@7/expression/left/right")` as "starting from root, resolve this path". I'm ambiguous to it though. It's fine with me to keep it. I usually like to keep the "stable" parameters to the left side of the list, since `path` changes more often than `root` (for example, when running in a loop).

spoenemann: 👍 "stable parameters to the left" is consistent with the context package.

path.go
sailingKieler: changed.

path.go
sailingKieler: fixed. (a plural 's' on `instance` was missing, removed ~instance of~)

examples/arithmetics/ast_locating_test.go
sailingKieler: Added one in `ast_locating_benchmark_test.go`, gives some interesting insight I can confirm.
Optimized the slice initializations/heap memory allocations a bit based on that.

path.go
sailingKieler: I don't see/feel that one arg is more stable than the other. When resolving reference targets the context root node has to be determined each time individually, anyway, since it could be defined in some other file.

In my (current) mind `path` is the primary info, while `node` is some kind of context ("Geraffel").
Maybe I'm still biased by https://github.com/cloudsmith/org.eclipse.emf/blob/0426d3bd36353798a212cb859ea4b5fbef301826/plugins/org.eclipse.emf.ecore/src/org/eclipse/emf/ecore/util/EcoreUtil.java#L254-L258

msujew: We did it the opposite way in Langium ;)
https://github.com/eclipse-langium/langium/blob/6c0e458edeb17f29ee09ceabb2ad8be26c52d1ed/packages/langium/src/workspace/ast-node-locator.ts#L32

sailingKieler: That doesn't count, as you did it that way according to your argumentation above.
Nevermind. I reversed them.

PR #125, opened 2026-07-01
Add lsp stub services

### LINKED ISSUE #28: Stub services for various LSP features
As part of the framework, we can provide some LSP features out of the box. However, not all of the LSP features can be implemented in this way. The framework should provide stubs similar to Langium's to help adopters write their semantic LSP services.

* [ ] Call Hierarchy
* [ ] Code Actions
* [ ] Code Lens
* [ ] Declaration
* [ ] Semantic Tokens
* [ ] Document Links
* [ ] Commands
* [ ] Hover
* [ ] Implementation
* [ ] Inlay Hint
* [ ] Signature Help
* [ ] Type Hierarchy
* [ ] Type

ssmifi:
Closes https://github.com/TypeFox/fastbelt/issues/28

Mostly ports from langium with go-idiomatic mechanisms for customization.

I refactored the original proposal. With the exceptions of Declaration, Implementation and Type Defintion, which provide meaningful defaults, all lsp stubs only provide the plain interface. 
Semantic tokens is another exception, which I kept for comparison, but I guess it will be overwritten by the more sophisticated version in #136. 

Added examples for Call Hierarchy, Code lens, Semantic tokens (which will be replaced), and Inlay Hints to the statemachine example to show adopters a working example. 

| # | Feature | Pattern | Notes |
|---|---|---|---|
| 1 | **Declaration** | Interface-only | Default removed - duplicated Definition without adding value |
| 2 | **Implementation** | Interface-only | Default + `ImplementationFilter` removed - the "type implements" heuristic requires language-specific type-system knowledge no generic filter can provide |
| 3 | **Type Definition** | Interface-only | Default removed - its cross-reference heuristic never actually filtered by field  |
| 4 | **Semantic Tokens** | — | Removed from this PR entirely; deferred to #136 |
| 5 | **Inlay Hint** | Interface-only | Unchanged. Open question: should it get `ResolvingInlayHintProvider` too? |
| 6 | **Signature Help** | Interface-only, + separate trigger service | Split into `SignatureHelpProvider`  and `SignatureHelpTriggers` |
| 7 | **Call Hierarchy** | Interface-only | Fixed `NodeAtCursor` - now uses `NameFinder`. Fixed in both the godoc example and the statemachine example |
| 8 | **Type Hierarchy** | Interface-only | Same `NodeAtCursor` |
| 9 | **Code Actions** | Interface-only, + optional resolve extension | Added `ResolvingCodeActionProvider` for `codeAction/resolve` support |
| 10 | **Code Lens** | Interface-only, + optional resolve extension | Added `ResolvingCodeLensProvider` for `codeLens/resolve` support. |
| 11 | **Document Links** | Interface-only | Unchanged. Open question: should it get `ResolvingDocumentLinkProvider` too? |
| 12 | **Commands** | Interface-only | Unchanged |


server/declaration_provider.go
```diff
@@ -0,0 +1,46 @@
...
+type DeclarationProvider interface {
+	HandleDeclarationRequest(ctx context.Context, params *lsp.DeclarationParams) ([]lsp.DefinitionLink, error)
+}
+
+// DefaultDeclarationProvider is the default implementation of [DeclarationProvider].
+// Language adopters should override this service if their language distinguishes
+// between declaration and definition.
+type DefaultDeclarationProvider struct {
+	sc *service.Container
+}
```
msujew: Suggestion: I would not provide a default implementation for this, since it exposes the "Go to Declaration" menu in the UI, which simply does the same thing as "Go to Definition" without providing any additional value.

server/implementation_provider.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// DefaultImplementationProvider is the default implementation of [ImplementationProvider].
+//
+// The default implementation finds all references to the target symbol and uses an
+// [ImplementationFilter] to determine which references represent actual implementations.
+type DefaultImplementationProvider struct {
+	sc     *service.Container
+	filter ImplementationFilter
+}
```
msujew: Discussion: I'm unsure how much benefit such a default/abstract implementation has, since it requires a very specific language design for this to work as expected.

server/inlay_hint_provider_test.go
```diff
@@ -0,0 +1,148 @@
...
+	assert.NoError(t, err)
+	assert.Empty(t, result, "server.NodesInRange should exclude declarations outside the requested range")
+}
+
+// grammarInlayHintProvider provides custom inlay hints for grammar language nodes.
+// Provider-only pattern: adopter implements the full provider interface,
+// using the shared server.NodesInRange helper for range-filtered iteration.
+type grammarInlayHintProvider struct {
+	sc *service.Container
+}
```
msujew, line 113: Question: Is this a real implementation or just for testing purposes?

server/signature_help_provider.go
```diff
@@ -0,0 +1,61 @@
...
+type SignatureHelpProvider interface {
+	// TriggerCharacters returns the characters that should trigger signature help.
+	TriggerCharacters() []string
+
+	// RetriggerCharacters returns the characters that should re-trigger signature help
+	// when it's already active.
+	RetriggerCharacters() []string
+
+	HandleSignatureHelpRequest(ctx context.Context, params *lsp.SignatureHelpParams) (*lsp.SignatureHelp, error)
+}
```
msujew, line 48: Suggestion: Should be split into two interfaces, see also how this is resolved for the completion triggers.

server/signature_help_provider_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+}
+
+// grammarSignatureHelpProvider provides custom signature help for grammar
+// language nodes. Provider-only pattern: adopter implements the full
+// provider interface, using the shared server.NodeAtCursor helper for
+// cursor-to-node resolution.
+type grammarSignatureHelpProvider struct {
+	sc *service.Container
+}
```
msujew, line 82: Question: Real or only for testing?

server/type_definition_provider.go
```diff
@@ -0,0 +1,88 @@
...
+
+// TypeDefinitionProvider is a service for handling LSP type definition requests.
+type TypeDefinitionProvider interface {
+	HandleTypeDefinitionRequest(ctx context.Context, params *lsp.TypeDefinitionParams) ([]lsp.DefinitionLink, error)
+}
+
+// DefaultTypeDefinitionProvider is the default implementation of [TypeDefinitionProvider].
+// The default implementation provides basic support by looking for type references
+// in the AST node's cross-references. Language adopters may need to override this
+// service to handle language-specific type inference or implicit type relationships.
```
msujew: Question: How does overriding this work? It only has a single method?

server/type_definition_provider.go
```diff
@@ -0,0 +1,88 @@
...
+		// Check if this is a type reference (many languages have a "type" feature)
+		// This is a heuristic and may need language-specific customization
+		targetDesc := xref.Description()
+		if targetDesc != nil && targetDesc.Node != nil {
+			targetNode := targetDesc.Node
+			targetTextDoc := targetNode.Document().TextDoc
+
+			// Check if the target looks like a type definition
+			// (This is a simple heuristic; language-specific logic would be better)
+			fullRange := targetNode.TextRange().LspRange(targetTextDoc)
```
msujew: Question: What is the heuristic here?

server/type_definition_provider.go
```diff
@@ -0,0 +1,88 @@
...
+	HandleTypeDefinitionRequest(ctx context.Context, params *lsp.TypeDefinitionParams) ([]lsp.DefinitionLink, error)
+}
+
+// DefaultTypeDefinitionProvider is the default implementation of [TypeDefinitionProvider].
+// The default implementation provides basic support by looking for type references
+// in the AST node's cross-references. Language adopters may need to override this
+// service to handle language-specific type inference or implicit type relationships.
+type DefaultTypeDefinitionProvider struct {
+	sc *service.Container
+}
```
msujew: Suggestion: I don't think we need a default implementation of this.

server/type_hierarchy_provider.go
```diff
@@ -0,0 +1,53 @@
...
+//
+//	type MyTypeHierarchyProvider struct{ sc *service.Container }
+//
+//	func (p *MyTypeHierarchyProvider) HandlePrepareTypeHierarchyRequest(ctx context.Context, params *lsp.TypeHierarchyPrepareParams) ([]lsp.TypeHierarchyItem, error) {
+//	    documentManager := service.MustGet[workspace.DocumentManager](p.sc)
+//	    doc := documentManager.Get(core.ParseURI(string(params.TextDocument.URI)))
+//	    if doc == nil {
+//	        return nil, nil
+//	    }
+//	    node := server.NodeAtCursor(doc, params.Position)
```
msujew: Suggestion: I believe this example is misleading. If you call `server.NodeAtCursor`, you run into two issues:
1. If the token at the cursor is a reference, you would expect that it shows the type hierarchy for the referenced type. However, `NodeAtCursor` will return the AST node that contains the reference.
2. If your cursor is over a random keyword of that type, like `t<cursor>ype SomeType`, it shows the type hierarchy for that type. Normally, it should only show it, if the user hovers over the name of the type (or a reference to the name).

server/code_lens_provider.go
```diff
@@ -0,0 +1,33 @@
...
+//	            Command: &lsp.Command{
+//	                Title:   "3 references",
+//	                Command: "showReferences",
+//	            },
+//	        },
+//	    }, nil
+//	}
+type CodeLensProvider interface {
+	HandleCodeLensRequest(ctx context.Context, params *lsp.CodeLensParams) ([]lsp.CodeLens, error)
+}
```
msujew, line 33: Suggestion: Can we also add a `ResolvingCodeLensProvider` that supports the [`codeLens/resolve`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLens_resolve) request?

server/code_action_provider.go
```diff
@@ -0,0 +1,31 @@
...
+//	        {
+//	            Title: "Add missing import",
+//	            Kind:  lsp.QuickFix,
+//	            Edit:  &lsp.WorkspaceEdit{...},
+//	        },
+//	    }, nil
+//	}
+type CodeActionProvider interface {
+	HandleCodeActionRequest(ctx context.Context, params *lsp.CodeActionParams) ([]lsp.CodeAction, error)
+}
```
msujew, line 31: Suggestion: Here as well for the [`codeAction/resolve`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction_resolve) request.

server/semantic_tokens_contributor.go
msujew, line 1: Suggestion: Not needed, see #136.

msujew (commented):
Thanks, that's very useful!

server/implementation_provider.go
ssmifi: I agree. Let's fall back to interface only.

server/inlay_hint_provider_test.go
ssmifi: It is a working implementation currently only used in testing. The "provider-only" comment was misleading though. I changed that already.

server/signature_help_provider_test.go
ssmifi: Same as before, working implementation currently only used for testing.

server/type_definition_provider.go
ssmifi: It is using the service override now `service.Put[server.TypeDefinitionProvider](sc, &grammarTypeDefinitionProvider{sc: sc})`. 
Depending on your thoughts about the other working test examples, we could keep the type definition provider test  as documentation. wdyt?

server/type_definition_provider.go
ssmifi: Superseded. Default is removed.

server/code_action_provider.go
ssmifi: Should we add this also for `InlayHint` and `DocumentLinks`?

ssmifi:
Thanks for the review @msujew. Included all suggestions. 
Still under consideration:
- Should we also add the resolve support pattern to the `InlayHintProvider` and `DocumentLinkProvider`?
- Whether to keep the real-but-test-only fixture providers (`grammarInlayHintProvider`, `grammarCodeLensProvider`, `grammarSignatureHelpProvider`, `grammarTypeDefinitionProvider`) as documentation, or trim them?

msujew:
> Should we also add the resolve support pattern to the InlayHintProvider and DocumentLinkProvider?

Good idea 👍 

> Whether to keep the real-but-test-only fixture providers (grammarInlayHintProvider, grammarCodeLensProvider, grammarSignatureHelpProvider, grammarTypeDefinitionProvider) as documentation, or trim them?

I think we can remove them. It doesn't feel like they actually do anything beyond asserting that the interface works as intended - which isn't doing a lot to be honest.

server/type_definition_provider.go
msujew: > Depending on your thoughts about the other working test examples, we could keep the type definition provider test as documentation. wdyt?

I don't think that's necessary, to be honest.

ssmifi:
Ok, implemented. Please have another look.

msujew:
FYI this needs a rebase due to LSP 3.18 update conflicts. I'll have another look once this is done.

server/server.go
```diff
@@ -61,6 +61,10 @@ func (s *DefaultLanguageServer) Initialize(ctx context.Context, params *lsp.Para
 	if service.Has[RenameProvider](s.sc) {
 		renameProvider = &lsp.RenameOptions{PrepareProvider: true}
 	}
+	var executeCommandProvider *lsp.ExecuteCommandOptions
+	if service.Has[CommandProvider](s.sc) {
+		executeCommandProvider = &lsp.ExecuteCommandOptions{Commands: []string{}}
+	}
```
msujew: Suggestion: The language server needs to publish which commands are actually available. Add this to the interface.

server/command_provider.go
```diff
@@ -0,0 +1,30 @@
...
+//	    case "myLanguage.refactor":
+//	        // Execute refactoring with params.Arguments
+//	        return "Refactoring completed", nil
+//	    default:
+//	        return nil, fmt.Errorf("unknown command: %s", params.Command)
+//	    }
+//	}
+type CommandProvider interface {
+	HandleExecuteCommandRequest(ctx context.Context, params *lsp.ExecuteCommandParams) (any, error)
+}
```
msujew, line 37: Suggestion: The language server needs to publish which commands are actually available. Add this to the interface.

server/server.go
```diff
@@ -84,10 +84,72 @@ func (s *DefaultLanguageServer) Initialize(ctx context.Context, params *lsp.Para
...
+				}
+				_, resolving := provider.(ResolvingDocumentLinkProvider)
+				return &lsp.DocumentLinkOptions{ResolveProvider: resolving}
+			}(),
+			ExecuteCommandProvider: func() *lsp.ExecuteCommandOptions {
+				if service.Has[CommandProvider](s.sc) {
+					return &lsp.ExecuteCommandOptions{Commands: []string{}}
+				}
+				return nil
+			}(),
```
msujew, line 131: Commands need to be published, see comment above.

server/signature_help_triggers.go
```diff
@@ -0,0 +1,28 @@
...
+	return &DefaultSignatureHelpTriggers{}
+}
+
+func (*DefaultSignatureHelpTriggers) TriggerCharacters() []string {
+	return nil
+}
+
+func (*DefaultSignatureHelpTriggers) RetriggerCharacters() []string {
+	return nil
+}
```
msujew: Suggestion: This implementation by itself doesn't really help adopters IMO. A much better implementation would be:

```go
func NewSignatureHelpTriggers(trigger, retrigger []string) SignatureHelpTriggers {
  return &signatureHelpTriggers{...} // <-- not exported default impl
}
```

server/server.go
```diff
@@ -152,91 +214,159 @@ func (s *DefaultLanguageServer) DidSave(ctx context.Context, params *lsp.DidSave
...
+}
+func (s *DefaultLanguageServer) ResolveCodeAction(ctx context.Context, params *lsp.CodeAction) (*lsp.CodeAction, error) {
+	provider, err := service.Get[CodeActionProvider](s.sc)
 	if err != nil {
-		return nil, err
+		return params, nil
+	}
+	resolving, ok := provider.(ResolvingCodeActionProvider)
+	if !ok {
+		return params, nil
```
msujew: Suggestion: Can we have a `dispatchResolveRead[Provider, ResolvingProvider, Result]` method for this? I think this can be handled generically.

msujew (commented):
Thanks, looks quite nice. Just a few remarks, see below.

server/server.go
ssmifi: Sure. I was on edge about this, but following our idoimatic go skill, claude actually recommended to keep these. Thinking about this longer, I would also tend to add a `dispatchResolveRead`.

msujew: I think it makes sense to use it. I'm fine with non-idiomatic Go, if it's in a non-exported context ;)

msujew (approved):
Thanks, I think this is good now.

PR #126, opened 2026-07-08
Reduce the amount of goroutines used for builds

msujew:
Massively reduces the amount of spawned goroutines required for a workspace build. I noticed that during large workspace builds, we spawn an unreasonable number of goroutines, which (a) drags down our memory profile and (b) slows down runtime performance when multithreading (but also when single-threading!).

This change adds a new `parallel.ForEachIter` (`iter.Seq[T]`) and `parallel.ForEach` (`[]T`) for efficient chunking of data for goroutines. This is now used for both the linking and general workspace build phase (essentially all steps where we use `sync.WaitGroup`), which results in a total speedup of roughly 40%, and a 20% memory usage decrease.

### DIFF
--- a/ast.go
+++ b/ast.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"sync/atomic"
 	"unique"
+
+	"typefox.dev/fastbelt/util/parallel"
 )
 
 // AstNode is the base interface for all AST nodes.
@@ -354,14 +356,27 @@ func MergeTokens(newNode AstNode, oldTokens []*Token) {
 	}
 }
 
-// AssignContainers recursively assigns document and parent pointers for root and its subtree.
+// Allocate a new reference slot for every ~10 tokens on average.
+// This average is updated after each traversal to adapt to the actual language.
+const defaultReferenceRatio = 1.0 / 10.0
+
+// running exponential moving average of references-per-token
+var avgReferenceRatio = parallel.NewRunningAverage(defaultReferenceRatio)
+
+// AssignContainers recursively assigns document and parent pointers for the root node and its subtree.
 //
 // It also assigns document and container on composite reference units reachable via references.
 // It will also fill the [Document.References] field with all references found in the subtree.
-func AssignContainers(doc *Document, root AstNode) {
-	references := []UntypedReference{}
-	doAssignContainers(doc, root, &references)
+func AssignContainers(doc *Document) {
+	// Continually increasing the capacity of the references slice is expensive
+	// So we use a running average of references-per-token to preallocate the slice
+	references := make([]UntypedReference, 0, avgReferenceRatio.Capacity(len(doc.Tokens)))
+	doAssignContainers(doc, doc.Root, &references)
 	doc.References = references
+
+	if len(doc.Tokens) > 0 {
+		avgReferenceRatio.Update(float64(len(references)) / float64(len(doc.Tokens)))
+	}
 }
 
 func doAssignContainers(doc *Document, root AstNode, references *[]UntypedReference) {
--- a/ast.go
+++ b/ast.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"iter"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"unique"
 
@@ -360,14 +361,27 @@ func MergeTokens(newNode AstNode, oldTokens []*Token) {
 // This average is updated after each traversal to adapt to the actual language.
 const defaultReferenceRatio = 1.0 / 10.0
 
-// running exponential moving average of references-per-token
-var avgReferenceRatio = parallel.NewRunningAverage(defaultReferenceRatio)
+// running exponential moving average of references-per-token for each language ID
+// different languages may have different average reference ratios
+var avgReferenceRatioMap sync.Map
+
+func getAvgReferenceRatio(languageId string) *parallel.RunningAverage {
+	if avg, ok := avgReferenceRatioMap.Load(languageId); ok {
+		return avg.(*parallel.RunningAverage)
+	} else {
+		avg := parallel.NewRunningAverage(defaultReferenceRatio)
+		avgReferenceRatioMap.Store(languageId, avg)
+		return avg
+	}
+}
 
 // AssignContainers recursively assigns document and parent pointers for the root node and its subtree.
 //
 // It also assigns document and container on composite reference units reachable via references.
 // It will also fill the [Document.References] field with all references found in the subtree.
 func AssignContainers(doc *Document) {
+	languageId := doc.TextDoc.LanguageID()
+	avgReferenceRatio := getAvgReferenceRatio(languageId)
 	// Continually increasing the capacity of the references slice is expensive
 	// So we use a running average of references-per-token to preallocate the slice
 	references := make([]UntypedReference, 0, avgReferenceRatio.Capacity(len(doc.Tokens)))
--- a/ast.go
+++ b/ast.go
@@ -357,14 +357,22 @@ func MergeTokens(newNode AstNode, oldTokens []*Token) {
 // AssignContainers recursively assigns document and parent pointers for root and its subtree.
 //
 // It also assigns document and container on composite reference units reachable via references.
+// It will also fill the [Document.References] field with all references found in the subtree.
 func AssignContainers(doc *Document, root AstNode) {
+	references := []UntypedReference{}
+	doAssignContainers(doc, root, &references)
+	doc.References = references
+}
+
+func doAssignContainers(doc *Document, root AstNode, references *[]UntypedReference) {
 	root.SetDocument(doc)
 	root.ForEachNode(func(child AstNode, containerField unique.Handle[string], index int) {
 		child.SetDocument(doc)
 		child.SetContainer(root, containerField, index)
-		AssignContainers(doc, child)
+		doAssignContainers(doc, child, references)
 	})
 	root.ForEachReference(func(ur UntypedReference, containerField unique.Handle[string], index int) {
+		*references = append(*references, ur)
 		unit := ur.Unit()
 		if stringNode, ok := unit.(CompositeNode); ok {
 			stringNode.SetDocument(doc)
--- a/util/parallel/parallel.go
+++ b/util/parallel/parallel.go
@@ -30,17 +30,20 @@ func ForEachIter[T any](seq iter.Seq[T], action func(T, int)) {
 // The second argument to action is the zero-based index of the element.
 // This function blocks until every action call has returned.
 func ForEach[T any](elements []T, action func(T, int)) {
-	if len(elements) == 0 {
+	total := len(elements)
+	if total == 0 {
 		return
 	}
-	workers := min(runtime.GOMAXPROCS(0), len(elements))
+	workers := min(runtime.GOMAXPROCS(0), total)
 	var next atomic.Int64
+	// Store -1 so that the first Add(1) returns 0, the index of the first element.
+	next.Store(-1)
 	var wg sync.WaitGroup
 	for range workers {
 		wg.Go(func() {
 			for {
-				i := int(next.Add(1) - 1)
-				if i >= len(elements) {
+				i := int(next.Add(1))
+				if i >= total {
 					return
 				}
 				action(elements[i], i)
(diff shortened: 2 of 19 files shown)


msujew:
Ignore comment above by the git-bot. It doesn't understand that 11MB/s is better than 5MB/s 🙄

util/parallel/parallel.go
```diff
@@ -0,0 +1,50 @@
...
+//
+// The second argument to action is the zero-based index of the element.
+// This function blocks until every action call has returned.
+func ForEach[T any](elements []T, action func(T, int)) {
+	if len(elements) == 0 {
+		return
+	}
+	workers := min(runtime.GOMAXPROCS(0), len(elements))
+	var wg sync.WaitGroup
+	chunk := (len(elements) + workers - 1) / workers
```
Lotes: On first sight I have no idea what you are doing here.
What is the meaning of `chunk`?
Are you do some kind of scheduling here? Big question marks...

Lotes: Ah it is some kind of scheduling (assigning tasks to workers)... but how did you get these formulas?
Some explanation would be nice. Ascii art could help haha, but I think words can serve the same purpose ^^*

util/parallel/parallel_test.go
```diff
@@ -0,0 +1,39 @@
...
+			}
+		})
+		for i, count := range seen {
+			if count != 1 {
+				t.Errorf("expected index %d to be visited once, got %d", i, count)
+			}
+		}
+	})
+
+	t.Run("empty sequence", func(t *testing.T) {
```
Lotes, line 48: Following tests come into my mind:

If the count of elements is:
* `== runtime.GOMAXPROCS(0)`
* `> runtime.GOMAXPROCS(0)`
* `< runtime.GOMAXPROCS(0)`
* `== 0`

Just to make sure no edge cases destroy the execution.

Lotes (commented):
Can you adjust the performance check in that regard that improvements of the speed are not flagged?
On the other hand this can also be an indicator that important code was skipped. But actually then the tests should also fail...

I added some comments about the parallelization module.

util/parallel/parallel.go
msujew: Not really scheduling, but rather parallel splitting of work into chunks/batches (one chunk per logical system thread). I.e. if you have 4 logical threads on your system, and you have to do work on 100 elements, each chunk has 25 elements, which are executed in order. I.e. first batch runs the given function on the first 25 elements, next batch on the 25-50 lements, and so on.

I'll add a comment to explain this.

util/parallel/parallel.go
```diff
@@ -34,9 +34,12 @@ func ForEach[T any](elements []T, action func(T, int)) {
 	}
 	workers := min(runtime.GOMAXPROCS(0), len(elements))
 	var wg sync.WaitGroup
-	chunk := (len(elements) + workers - 1) / workers
+	// Compute the amount of batches to process
+	// Note that this is not the same as the amount of workers,
+	// since the amount of workers may be larger than the amount of elements.
+	batchCount := (len(elements) + workers - 1) / workers
```
Lotes: `batchSize` or `batchElementCount`?
`batchCount` reads like the count of batches (==#workers actually)

util/parallel/parallel.go
Lotes: What if you have 100 elements and 2 workers and the first 50 take a long amount of time and the last 50 are pretty fast? The second worker is done and idling while the first one is still processing...

An alternative could be a queue with an cheap lock/sync mechanism in front of it.

But that is all so theoretical. You could find out with a benchmark. Or maybe you already have an answer to this?

util/parallel/parallel_test.go
Lotes: Mh it is a bit of a disadvantage that this in concurrently. Harder to test.

It might make more sense to test the correct assignment of elements to workers.

But slowly I understand the mechanism, it is not so complicated, actually...

util/parallel/parallel.go
msujew: The queue (with locking) is also rather expensive. There is no cheap locking if you want to push 100k+ elements through 16 cores. This is a trade-off, which in practice, should fare rather well.

Note that modern systems all have 8+ logical cores, meaning that this effect is less pronounced, since each individual slice is smaller.

util/parallel/parallel.go
msujew: Good point, thanks!

util/parallel/parallel_test.go
msujew: > It might make more sense to test the correct assignment of elements to workers.

That sounds rather brittle. The function is an abstraction over "do things in parallel, I don't care how". I'm not even sure whether you have enough runtime information in Go to actually know which worker executed what. There is no `threadId` or similar as you would have when doing actual multi-threading.

util/parallel/parallel.go
Lotes: Ok. A mixture is also inventable: a queue of batches of a fixed size. But as long as we don't know whether we have a problem with the current approach...

util/parallel/parallel.go
```diff
@@ -0,0 +1,53 @@
...
+	ForEach(slices.Collect(seq), action)
+}
+
+// ForEach calls action once for each element in the given slice,
+// distributing the calls across at most [runtime.GOMAXPROCS](0) goroutines
+// instead of one goroutine per element.
+//
+// The second argument to action is the zero-based index of the element.
+// This function blocks until every action call has returned.
+func ForEach[T any](elements []T, action func(T, int)) {
```
Lotes, line 32: Question: Is it not normal to also pass a context to concurrent tasks?

util/parallel/parallel.go
msujew: Nevermind, the queue actually seems to have the same performance, I will use that instead.

util/parallel/parallel_test.go
Lotes: Ah misunderstanding! I meant you could extract the assignment function "somehow" without using anything thread-like. Just a function. But as I said, the formula makes sense to me now...

util/parallel/parallel.go
msujew: Not really. If the action requires it, it will use the outer context. Since `ForEach` doesn't generate its own context, an outer context is good enough.

util/parallel/parallel_test.go
Lotes: Obsolete now.

Lotes (approved):
Fine for me.

ast.go
```diff
@@ -357,14 +357,22 @@ func MergeTokens(newNode AstNode, oldTokens []*Token) {
 // AssignContainers recursively assigns document and parent pointers for root and its subtree.
 //
 // It also assigns document and container on composite reference units reachable via references.
+// It will also fill the [Document.References] field with all references found in the subtree.
 func AssignContainers(doc *Document, root AstNode) {
+	references := []UntypedReference{}
```
sailingKieler: Considering this side effect I suggest to remove the func parameter `root` and refer to `doc.Root` below,
since now calling `AssignContainers` with an AstNode `!= doc.Root` most likely leads to errors.

workspace/builder.go
sailingKieler, line 207: Proposal:
Change case distinction to `switch` statement with explicit `fallthrough`, like:
```
switch {
case !state.Has(core.DocStateParsed):
		doc.Root = nil
		doc.Tokens = core.TokenSlice{}
		doc.ParserErrors = []*core.ParserError{}
		doc.LexerErrors = []*core.LexerError{}
		doc.References = []core.UntypedReference{}
		fallthrough
case !state.Has(core.DocStateExportedSymbols):
...
}
```
This avoid errors due to crude values of `state`, like `core.DocStateParsed | core.DocStateLinked`, and avoid the pre-treatment of the references.
All steps following the first negative flag will be applied.

In fact, the current/previous impl has a bug:
If a reference in the doc has been resolved transitively before the document is actually linked (reached `core.DocStateLinked`) those references would not been reset by `Reset(...)`.

util/parallel/parallel.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	workers := min(runtime.GOMAXPROCS(0), len(elements))
+	// Note: we use a channel to distribute indices to workers
+	// This ensures that there is no idle worker while there are still elements to process
+	indices := make(chan int, workers)
+	var wg sync.WaitGroup
+	for range workers {
+		wg.Go(func() {
+			for index := range indices {
+				action(elements[index], index)
```
sailingKieler: This shared array variable looks suspicious to me, although it's never updated.

What about sharing references/pointers to the items directly to the go routines via the channel?
Do we actually need the index during parallel processing in common cases?

sailingKieler (commented):
Nice improvement @msujew 👍
I have some remarks and suggestions, see below.

workspace/builder.go
msujew: > In fact, the current/previous impl has a bug:
If a reference in the doc has been resolved transitively before the document is actually linked (reached core.DocStateLinked) those references would not been reset by Reset(...).

I'm not sure I follow. We don't even look at the document state when resetting the references? `state` is the target state, not the actual document state.

I'll implement the `switch-case` proposal though.

util/parallel/parallel.go
msujew: I've refactored this, to use a channel-less design. It's much faster now.

msujew:
FYI I've also introduced a `parallel.RunningAverage` struct that helps us reduce effort while populating the `References` slice. It's reused in the lexer as well.

workspace/builder.go
sailingKieler: > I'm not sure I follow. We don't even look at the document state when resetting the references? state is the target state, not the actual document state.

You're right, the flag/bit-based indication still confuses me sometimes.
We had this issue with Langium once, but there was an early exit (doc.state < ...) at that time.

util/parallel/parallel.go
```diff
@@ -33,20 +34,18 @@ func ForEach[T any](elements []T, action func(T, int)) {
...
-	// This ensures that there is no idle worker while there are still elements to process
-	indices := make(chan int, workers)
+	var next atomic.Int64
 	var wg sync.WaitGroup
 	for range workers {
 		wg.Go(func() {
-			for index := range indices {
-				action(elements[index], index)
+			for {
+				i := int(next.Add(1) - 1)
```
sailingKieler: Would it work to initialize next with `-1` and avoid the subtraction here?

util/parallel/parallel.go
```diff
@@ -33,20 +34,18 @@ func ForEach[T any](elements []T, action func(T, int)) {
...
+				if i >= len(elements) {
```
sailingKieler: Would prefetching `len(elements)` into a variable make any notable difference?

workspace/builder.go
```diff
@@ -172,36 +172,33 @@ func (s *DefaultBuilder) Build(ctx context.Context, docs []*core.Document, downg
...
+		fallthrough
+	case !state.Has(core.DocStateImportedSymbols):
 		doc.ImportedSymbols = nil
-	}
-	if !state.Has(core.DocStateLocalSymbols) {
+		fallthrough
+	case !state.Has(core.DocStateLocalSymbols):
 		doc.LocalSymbols = nil
-	}
-	if !state.Has(core.DocStateReferences) {
```
sailingKieler, line 202: missing `fallthrough` here and below?

ast.go
```diff
@@ -354,14 +356,27 @@ func MergeTokens(newNode AstNode, oldTokens []*Token) {
 	}
 }
 
-// AssignContainers recursively assigns document and parent pointers for root and its subtree.
+// Allocate a new reference slot for every ~10 tokens on average.
+// This average is updated after each traversal to adapt to the actual language.
+const defaultReferenceRatio = 1.0 / 10.0
+
+// running exponential moving average of references-per-token
+var avgReferenceRatio = parallel.NewRunningAverage(defaultReferenceRatio)
```
sailingKieler: In a multi-language context we should think about distinguishing this ratio per language, shouldn't we?

sailingKieler (approved):
This is basically fine for me, but I have some additional findings, see below.

Please update the PR title before merging 😉

sailingKieler (approved):
Great, good to go from my point of view! 🚀

sailingKieler:
... don't forget to update the title/final commit msg 😉

PR #128, opened 2026-07-10
Use byte offsets for internal ranges

msujew:
To reduce on memory usage, this change removes the line/column information for all internal text ranges and only leaves the byte-offset ranges in place. On one hand, this simplifies some code, while making some other more complicated (especially LSP response handling).

Reduces the memory usage of tokens by 20% (16 bytes). Also removes 16 bytes from all AST nodes. Since we no longer have to track utf16 line/column infos, this speeds up the lexer by roughly 10%.


text.go
```diff
@@ -4,61 +4,23 @@
...
-	// End is the exclusive end byte offset.
-	End TextIndex
+import (
+	"typefox.dev/fastbelt/textdoc"
+	"typefox.dev/lsp"
+)
+
+// Range represents a text range in a document. The indices are zero-based and represent byte offsets.
+// It is defined by a start and end offset, where the start is inclusive and the end is exclusive.
+type Range struct {
```
sailingKieler: Why not keeping the name `TextRange`?
It would be a bit more explicit, and since it's defined in the root/core package there's no stuttering.

ast.go
```diff
@@ -40,21 +40,21 @@ type AstNode interface {
...
 	// It is primarily used by generated parsers while constructing nodes incrementally.
-	SetSegment(segment *TextSegment)
-	// SetSegmentStartToken sets the start of the node's segment from token.
+	SetRange(r Range)
+	// SetRangeStartToken sets the start of the node's range from token.
 	//
 	// It is primarily used by generated parsers while constructing nodes incrementally.
-	SetSegmentStartToken(token *Token)
-	// SetSegmentEndToken sets the end of the node's segment from token.
+	SetRangeStartToken(token *Token)
```
sailingKieler: Since the implementation of this and the following method is rather trivial now, is it still worth keeping it?
Looks like some `SetRangeStart(int32)` and `SetRangeEnd(int32)` would be sufficient.
(The names of the current methods feel somewhat confusing to me)

The `nil` checks could be moved to some non-member utility func, are they actually relevant in practice?

internal/generator/parser_generator.go
```diff
@@ -1114,11 +1114,11 @@ func generateRuleCallParser(node codegen.Node, context *ParserGeneratorContext,
 				n.AppendLine("p.state.ExitRule()")
 			} else {
 				n.AppendLine("result ", eq, " core.NewCompositeNode()")
-				n.AppendLine("result.SetSegmentStartToken(p.state.LA(1))")
+				n.AppendLine("result.SetRangeStartToken(p.state.LA(1))")
 				n.AppendLine("p.state.EnterRule(", followName, ")")
 				n.AppendLine("p.Parse", t.Name(), "(result)")
 				n.AppendLine("p.state.ExitRule()")
-				n.AppendLine("result.SetSegmentEndToken(p.state.LA(0))")
+				n.AppendLine("result.SetRangeEndToken(p.state.LA(0))")
```
sailingKieler: Could the assignment of the range start above postponed and combined with this assignment here using `SetRange(...)`?

sailingKieler (commented):
The change looks pretty good to me, and makes sense.
In practice, I assume only the workspace symbol and document symbol providers might notably be affected performance-wise. In critical cases we always can modify the concrete `...Impl` struct generation to mitigate potential runtime performance issues - if such ever show up, right?

I have just some API-related hints/questions.

text.go
spoenemann: Another thought: as the root `fastbelt` package already depends on `textdoc` (Document has a TextDoc field), WDYT of moving this file to `textdoc` so the exported elements read well without the "Text" prefix (e.g. `textdoc.Range`, `textdoc.Location` etc.)?

internal/generator/parser_generator.go
msujew: No, due to we implement actions, we need separate ways to store start/end ranges. Everything else would be too much of a hassle.

text.go
msujew: I tried that out, but I didn't like moving the data into a package that doesn't make any use of it. I've renamed it to `fastbelt.TextRange` and renamed everything else appropriately. I also like it now more, since AstNodes can now use `Range` as a property name, which isn't unusual.

ast.go
msujew: Not relevant anymore, `LA` never returns `nil` anymore as of `v0.1.0`.

test/doc_fixture.go
```diff
@@ -216,53 +213,40 @@ func (d *Doc) markerRanges(label string) []core.TextRange {
...
-			start := d.Document.TextDoc.PositionAt(r.Start)
-			end := d.Document.TextDoc.PositionAt(r.End)
-			result = append(result, core.TextRange{Start: core.TextLocation{
-				Line:   core.TextLine(start.Line),
-				Column: core.TextColumn(start.Character),
-			}, End: core.TextLocation{
-				Line:   core.TextLine(end.Line),
-				Column: core.TextColumn(end.Character),
-			}})
+			result = append(result, core.TextRange{
```
sailingKieler: This can be replaced, right?

```suggestion
			result = append(result, core.NewTextRange(r.start, r.end)
```

text_range.go
sailingKieler, line 1: Rename to `text_range.go`?

sailingKieler (approved):
This looks really nice, many source portions feel much more streamlined.

I have just two very minor points, see below. 🚀

PR #129, opened 2026-07-21
Package documentation for parser and server

spoenemann:
- Added parser/doc.go
- Added server/doc.go and added missing documentation of public elements
- Extended util/parallel/doc.go
- Added package doc references to AGENTS.md

### DIFF
--- a/parser/doc.go
+++ b/parser/doc.go
@@ -1,0 +1,121 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+// Package parser provides the shared parser runtime used by Fastbelt-generated
+// languages. It does not implement parsing for any particular language; each
+// grammar gets its own generated recursive-descent parser that drives the
+// parser state, adaptive prediction, error recovery, and completion machinery
+// of this package.
+//
+// Run [typefox.dev/fastbelt/cmd/fastbelt] generate on a .fb grammar file to
+// emit the files described below. See [typefox.dev/fastbelt] for the overall
+// toolchain and [typefox.dev/fastbelt/grammar] for how parser rules are
+// declared in the grammar language.
+//
+// # Generated parsers
+//
+// Code generation turns the parser rules of a grammar into four cooperating
+// files:
+//
+//   - `parser_gen.go` — a Parser implementing [Parser], with one ParseX method
+//     per parser rule. Each method allocates the rule's AST node, matches
+//     tokens through [ParserState], assigns them to node fields, and records
+//     the node's token segment for source ranges.
+//   - `parser_lookahead_gen.go` — the lookahead service, with one method per
+//     decision point in the grammar: which alternative to take, whether to
+//     enter an optional group, whether to iterate a loop once more. It is an
+//     interface with a default implementation, so individual decisions can be
+//     overridden without touching the parser.
+//   - `atn_gen.go` — the [RuntimeATN], an augmented transition network that
+//     encodes the grammar as a graph of states with token, epsilon, and rule
+//     transitions. It backs adaptive prediction, error recovery, and
+//     completion.
+//   - `completion_parser_gen.go` — a stripped-down twin of the parser that
+//     builds no AST and instead records the parse progress snapshots that
+//     code completion needs.
+//
+// The generated `SetupGeneratedServices` registers the Parser, the lookahead
+// service, and the default [ErrorRecoveryStrategy] and [ErrorMessageProvider]
+// in the service container ([typefox.dev/fastbelt/util/service]), each unless
+// a custom implementation was registered first. During document processing,
+// [typefox.dev/fastbelt/workspace.DefaultDocumentParser] obtains the [Parser],
+// calls Parse on the tokens produced by [typefox.dev/fastbelt/lexer], and
+// stores the resulting root [core.AstNode] and syntax errors ([ParseResult])
+// on the document.
+//
+// # Parsing model
+//
+// A generated parser is a recursive-descent parser: one method per rule,
+// descending into sub-rules and consuming tokens left to right. [ParserState]
+// carries the per-parse state: token slice, cursor, error mode, and rule stack.
+//
+// Wherever the grammar offers a choice — between alternatives, at an optional
+// element, at a loop continuation — the parser asks its lookahead service.
+// Depending on the grammar, the generated decision method uses one of three
+// strategies, cheapest first:
+//
+//  1. A direct check of the next token's type, when a single expected token
+//     decides.
+//  2. An [LL1Lookahead] table resolved by [ParserState.Lookahead] in O(1),
+//     when one token of lookahead decides.
+//  3. Adaptive ALL(*) prediction via [ParserState.AdaptivePredict], when
+//     arbitrary lookahead is required. The predictor simulates the
+//     [RuntimeATN] ahead of the current position and caches its decisions in
+//     per-decision DFAs, so hot decisions approach table-lookup cost over
+//     time. The cache is internally synchronized and shared across
+//     concurrent parses.
+//
+// Adaptive prediction runs in [PredictionModeSLL] by default;
+// [ParserLookahead.SetPredictionMode] enables full-context [PredictionModeLL]
+// for grammars that need it.
+//
+// Cross-reference assignments in the grammar do not resolve anything at parse
+// time: the parser stores an unresolved reference created by the generated
+// references constructor, and the [typefox.dev/fastbelt/linking] phase
+// resolves it later.
+//
+// # Error recovery
+//
+// Parse errors do not stop the parse. [Parser] implementations always return
+// an AST — with nil tokens or unset fields where input was missing — plus the
+// collected [core.ParserError] values, so language tooling keeps working on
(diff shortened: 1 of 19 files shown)


server/doc.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+// Package server implements the Language Server Protocol (LSP) side of
+// Fastbelt. It exposes the language services built by
+// [typefox.dev/fastbelt/workspace] — parsed, linked, and validated documents —
+// to editors and IDEs: text documents are synchronized through LSP
+// notifications, and language features such as completion, go to definition,
+// find references, hover, and rename are answered from the workspace.
```
sailingKieler, line 10: This sentence sounds a bit odd, e.g.
> ... are answered from the workspace.

server/doc.go
```diff
@@ -0,0 +1,84 @@
...
+// notification.
+//
+// A typical main function for a generated language looks like this:
+//
+//	func main() {
+//		sc := service.NewContainer()
+//		mylang.SetupServices(sc) // language, textdoc, linking, workspace services
+//		mylang.SetupGeneratedServerServices(sc)
+//		server.SetupDefaultServices(sc)
+//		server.SetupStdioServices(sc)
```
sailingKieler, line 74: Currently, we have it everywhere like this. However, intuitively I would expect the following pattern starting with basic/default definitions followed by the specific ones, esp. since there is a function `service.Override[...](...)` now.
```
		server.SetupDefaultServices(sc)
		server.SetupStdioServices(sc)
		mylang.SetupGeneratedServerServices(sc)
		mylang.SetupServices(sc) // language, textdoc, linking, workspace services
```
What do you think?

server/document_highlight_provider.go
```diff
@@ -13,14 +13,19 @@ import (
...
 }
 
+// DefaultDocumentHighlightProvider is the default implementation of [DocumentHighlightProvider].
+// It highlights all references to the symbol at the cursor position that are
+// located in the same document, using [NameFinder] and [ReferencesFinder].
 type DefaultDocumentHighlightProvider struct {
 	sc *service.Container
 }
 
+// NewDefaultDocumentHighlightProvider creates a new default document highlight provider.
```
sailingKieler, line 28: Nitpick, here and below:
```suggestion
// NewDefaultDocumentHighlightProvider creates a new [DefaultDocumentHighlightProvider].
```

server/server.go
```diff
@@ -26,7 +26,14 @@ type InitializeParticipant interface {
 	OnServerInitialize(params *lsp.ParamInitialize)
 }
 
-// DefaultLanguageServer implements the [lsp.Server] interface
+// DefaultLanguageServer is the default implementation of the [lsp.Server]
+// interface. It contains no feature logic itself: each supported LSP method
+// is delegated to a dedicated service (such as [CompletionProvider] or
+// [DocumentSyncher]) looked up in the service container, and request-handling
```
sailingKieler, line 32: The `DocumentSyncher` is a bit of an unfavorable example wrt. to the beginning of the sentence, `DefinitionProvider` or `DocumentSymbolProvider` are a bit better suited I think.

server/doc.go
```diff
@@ -0,0 +1,84 @@
...
+// registered with [SetupStdioServices] (communication over stdin/stdout) or
+// SetupWasmServices (communication with a JavaScript host, available in
+// js/wasm builds). It dials a JSON-RPC connection using the registered
+// [golang.org/x/exp/jsonrpc2.Dialer] and [golang.org/x/exp/jsonrpc2.Binder]
+// — by default [DefaultBinder], which attaches the [typefox.dev/lsp.Server]
+// from the container as the message handler — and blocks until the
+// connection is closed, which the default server triggers on the LSP exit
+// notification.
+//
+// A typical main function for a generated language looks like this:
```
sailingKieler, line 67:
```suggestion
// A typical main function of a language server looks like this:
```

server/doc.go
sailingKieler: At that I noticed that `mylang.SetupGeneratedServerServices(sc)` introduced within #79 is missing in `examples/arithmetics/server/main.go`, since it is also missing in the scaffold template in `internal/scaffold/templates/cmd/main.go.tmpl`.

Would you complete those gaps?

sailingKieler (commented):
Thanks a lot @spoenemann for these documentations.

I have a few (nitpick) remarks and found one gap (actually two) while reading the docs.

parser/doc.go
```diff
@@ -0,0 +1,121 @@
...
+//     code completion needs.
+//
+// The generated `SetupGeneratedServices` registers the Parser, the lookahead
+// service, and the default [ErrorRecoveryStrategy] and [ErrorMessageProvider]
+// in the service container ([typefox.dev/fastbelt/util/service]), each unless
+// a custom implementation was registered first. During document processing,
+// [typefox.dev/fastbelt/workspace.DefaultDocumentParser] obtains the [Parser],
+// calls Parse on the tokens produced by [typefox.dev/fastbelt/lexer], and
+// stores the resulting root [core.AstNode] and syntax errors ([ParseResult])
+// on the document.
```
msujew, line 45: Suggestion: Remove the whole last sentence here. It's not part of the generated parser infrastructure and I would like to change this soon, because I'm not happy with how this currently works. Having this documentation here would lead to me having more work in the future.

parser/doc.go
```diff
@@ -0,0 +1,121 @@
...
+// broken input. While the parser is recovering, follow-up errors are
+// suppressed so one underlying mistake produces one diagnostic. The pluggable
+// [ErrorRecoveryStrategy] controls resynchronization:
+//
+//   - [DefaultErrorRecovery] attempts single-token deletion when Consume hits
+//     an unexpected token, discards tokens before optional/loop guards until
+//     one fits the decision or an enclosing follow set (Sync), and skips to
+//     the follow set after unwinding from a failed rule (Recover). Because
+//     generated AST nodes tolerate nil tokens, missing tokens are reported
+//     but never fabricated.
```
msujew, line 92: I don't understand the last sentence. Maybe just say: "The error recovery never generates synthetic tokens for the purpose of a valid AST"?

parser/doc.go
```diff
@@ -0,0 +1,121 @@
...
+//     via [CompletionParserState] it records an [ATNSnapshot] — token index,
+//     ATN state, rule stack — at every rule entry and sync point.
+//  2. [CompletionParseResult.SimulateAt] picks a suitable snapshot and
+//     advances an NFA-style live set over the ATN ([RuntimeATN.Simulate])
+//     through the remaining tokens. Unlike prediction, the simulation keeps
+//     every alternative alive: the user has not committed to a branch yet.
+//  3. [RuntimeATN.NextCompletionsFromSet] reports what may come next as a
+//     [CompletionInfo]: valid token types, plus [CompletionHint] entries that
+//     mark cross-reference positions so the provider can offer resolvable
+//     symbols instead of a bare identifier token.
```
msujew, line 116: Suggestion: Too in depth? I'm not sure this is interesting for adopters. They should never need to touch this or understand it beyond the basic workings of the parser (described above).

msujew (commented):
I feel like this is rather wordy. I've picked some examples which I didn't like, but I feel like we would do well to cut down a bit. It explains too much of the inner workings, less of the API (which is the more interesting part to readers).

parser/doc.go
```diff
@@ -0,0 +1,121 @@
...
+//   - `parser_gen.go` — a Parser implementing [Parser], with one ParseX method
+//     per parser rule. Each method allocates the rule's AST node, matches
+//     tokens through [ParserState], assigns them to node fields, and records
+//     the node's token segment for source ranges.
+//   - `parser_lookahead_gen.go` — the lookahead service, with one method per
+//     decision point in the grammar: which alternative to take, whether to
+//     enter an optional group, whether to iterate a loop once more. It is an
+//     interface with a default implementation, so individual decisions can be
+//     overridden without touching the parser.
+//   - `atn_gen.go` — the [RuntimeATN], an augmented transition network that
```
Lotes, line 30: We could also mention the optional ATN markdown generator here. Its purpose was simpler debugging using mermaid diagrams.

server/doc.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+// Package server implements the Language Server Protocol (LSP) side of
```
Lotes, line 5: This package ...
The package 'server' ... (quotes)

Otherwise it reads as we have a PACKAGE SERVER.
Also the article should not be forgotten.

server/document_sync.go
```diff
@@ -72,6 +72,7 @@ type DefaultDocumentSyncher struct {
 	mu                        sync.RWMutex
 }
 
+// NewDefaultDocumentSyncher creates a new default document syncher.
 func NewDefaultDocumentSyncher(sc *service.Container) DocumentSyncher {
```
Lotes, line 76: Out of scope, but WDYT?
In my eyes "Syncher" is very informal or even bad English.
What about "Synchronizer" or "Reconciler" as alternative?

msujew: I like `Synchronizer` as well.

PR #127, opened 2026-07-10
Introduce token modes

Lotes:
Closes #50

* added explicit token modes (lexer modes)
* if none is given, one is create implicitly
* if not given, it collects all keywords, token declarations and token groups from the grammar
* token modes can have following members
  * own token declarations
  * token refs to top-level token declaration
  * own token groups
  * keywords
  * keyword selection using a RegExp
* token commands were also introduced
  * `push(name)` to push a token mode onto the token mode stack
  * `pop` to pop the topmost token mode
  * `mode(name)` to set the topmost token mode 
* I also extended token declarations to be keywords as well
* a lot of changes happened in the lexer generator
  * I tidied up the code organization a lot, I hope you like it as well!
  * token types are split from group type and token commands
    * it made sense during the journey in order to write a command/type differently in different modes... we can discuss if it still makes sense
  * "keywords" I renamed conceptually to `KeywordSelector`
* validations around token commands were created
* a lot of tests were added by AI, they caught bugs and show the expectations when using token modes

Issue #83, opened 2026-05-27
[generator] Support loading additional grammar files from same package


spoenemann:
See discussion in #81: The fastbelt CLI should load all .fb files in the same package as the specified entry file. Those shall be used to resolve references to interfaces, rules, etc.

All code generators should be able to handle a collection of grammar files rather than just one.

PR #130, opened 2026-07-23
Implement multi-language support

### LINKED ISSUE #96: [Improvement] Support multi-language setups
See also the discussion in [here](https://github.com/TypeFox/fastbelt/discussions/63#discussioncomment-16636210) for hints on how we could implement this.

msujew:
Closes https://github.com/TypeFox/fastbelt/issues/96

Does as the title says, similar to how it is outlined in the issue:
* Adds a `fastbelt/cmd` package adopters can use to write their own CLIs to generate Fastbelt code.
* Adds a `LanguageSelector` service that is used to determine during runtime what language a URI/document belongs to.
* Implements a switch-case based selector in the parser and lexer to circumvent the need for a separate service instance per language.
* Adds a `multilang` example that shows that this is working as expected (including parser, lexer and completion tests).

FYI, while this has a `Plugins` field in the API, plugins aren't actually supported in this PR. This will need a follow-up.


cdietrich:
wondering about the common lexical scope and an option to have a lexer per entry rule
https://github.com/TypeFox/fastbelt/pull/131

msujew:
@cdietrich Thanks for the input. I plan to finish implementing this once https://github.com/TypeFox/fastbelt/pull/127 lands, since it refactors a lot of the lexer infrastructure.

cmd/fastbelt/generate.go
```diff
@@ -55,7 +55,7 @@ func runGenerateCLI(opts generateOptions) error {
 	}
 
 	sc := grammar.CreateServices()
```
ssmifi, line 57: Nice separation. 
_Question_: The following lines are conceptionally doing for one file what `parseAndMerge` and `reportDiagnostic` do generically now. I was wondering if we can de-duplicate some code here and use come common helper?

cmd/fastbelt/generate.go
```diff
@@ -55,7 +55,7 @@ func runGenerateCLI(opts generateOptions) error {
...
-	file, _ := textdoc.NewFile(lsp.URIFromPath(grammarPath), "fb", 0, string(grammarText))
+	file := textdoc.NewFile(lsp.URIFromPath(grammarPath), "fb", 0, string(grammarText))
```
ssmifi, line 58: _Question_: So, while the API now allows for a whole directory of .fb files, the CLI just allows for one file?

cmd/build.go
```diff
@@ -0,0 +1,332 @@
...
+	"typefox.dev/fastbelt/internal/generator"
+	"typefox.dev/fastbelt/internal/grammar"
+	"typefox.dev/fastbelt/textdoc"
+	"typefox.dev/fastbelt/util/service"
+	"typefox.dev/fastbelt/workspace"
+	"typefox.dev/lsp"
+)
+
+// Language configures one language served by the generated language server: the
+// grammar entry rule it parses from and the [Selector] that claims its
```
ssmifi, line 29: _Minor remark_: I'm not sure about the go docs here. There is no type `Selector` in this package. The real type is `generator.Selector`. Just a hint if it is importannt for the docs or else. Besides, `Language` does not really hold a selector. `Build` constructs a `generator.Selector` from the `LanguageId`.

cmd/build.go
```diff
@@ -0,0 +1,332 @@
...
+
+// collectGrammarFiles returns the absolute paths of top-level *.fb files in dir.
+func collectGrammarFiles(dir string) ([]string, error) {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return nil, err
+	}
+	var files []string
+	for _, e := range entries {
+		if e.IsDir() || filepath.Ext(e.Name()) != ".fb" {
```
ssmifi, line 120: _Question_: So only .fb in this directory are considered? Grammars split across several subdirectories seems like a plausible future ask.

cmd/build.go
```diff
@@ -0,0 +1,332 @@
...
+	Plugins   []Plugin
+	ATN       bool
+	Verbose   bool
+}
+
+// Build parses and links every .fb file under Input, validates that each
+// language's entry rule exists and is marked entry, and generates the combined
+// language package into Output.
+func (c *BuildContext) Build() error {
+	for _, p := range c.Plugins {
```
ssmifi, line 59: _Question_: Since plugins are not supported yet, should we ship this here? Any substantial change later could break the API. Then again we are early in development and if you already have anythong concrete in mind, I guess it is also worth keeping.

examples/multilang/server/main.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2025 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+//go:build !js || !wasm
```
ssmifi: This does not look quite right here.

examples/multilang/server/main.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+	"log"
+
+	"typefox.dev/fastbelt/examples/multilang"
+	"typefox.dev/fastbelt/server"
+)
+
+func main() {
```
ssmifi: _Question_: The statemachine example also has a `wasm` entry point. Not necessary, but maybe you want to provide one here as well?

examples/multilang/completion_parser_gen.go
```diff
@@ -0,0 +1,72 @@
...
+// the ATN simulator.
+func (p *CompletionParser) Parse(document *core.Document, tokens []core.Token) *parser.CompletionParseResult {
+	messages := service.MustGet[parser.ErrorMessageProvider](p.sc)
+	recovery := service.MustGet[parser.ErrorRecoveryStrategy](p.sc)
+	lookahead := service.MustGet[MultilangModelParserLookahead](p.sc)
+	cp := &CompletionParser{sc: p.sc, atn: p.atn, lookahead: lookahead}
+	cp.state = parser.NewParserState(tokens, p.atn(), recovery, messages)
+	cp.cp = parser.NewCompletionParserState(cp.state)
+	selector := service.MustGet[core.LanguageSelector](cp.sc)
+	switch i, _ := selector.Select(core.ParseURI(string(document.TextDoc.URI()))); i {
```
ssmifi: _Question_: The `switch` in `parser_gen.go` does not rely on the `core.ParseURI`. Wouldn't `document.URI` not suffice here?

internal/generator/lexer_generator.go
```diff
@@ -142,27 +144,131 @@ func GenerateLexer(grammr grammar.Grammar, packageName string, tokenTypes Genera
...
+				}
+			}
+		}
+	}
+	visitedRules[entry.Name()] = true
+	visitRule(entry)
+
+	for _, token := range grammr.Terminals() {
+		if token.Type() == "hidden" || token.Type() == "comment" {
+			reachable[GeneratedTokenName(token)] = true
```
ssmifi, line 266: _Question_: So, every hidden token is always in the token list? So, is it/will it be possible in the future to combine languages with lets say different comment terminals, e.g., `// line comment` and `# line comment`?

internal/generator/services_generator.go
```diff
@@ -26,6 +33,26 @@ func GenerateServices(grammr grammar.Grammar, packageName string) string {
...
+			n.Indent(func(n codegen.Node) {
+				n.AppendLine("service.Put[core.LanguageSelector](")
+				n.Indent(func(n codegen.Node) {
+					n.AppendLine("sc,")
+					n.AppendLine("core.NewDefaultLanguageSelector(")
+					n.Indent(func(n codegen.Node) {
+						n.AppendLine("sc,")
+						for _, s := range selectors {
+							patterns := strings.Join(s.Patterns, "\", \"")
+							n.AppendLine("core.NewDocumentSelectorWithPatterns(\"", s.LanguageID, "\", \"", patterns, "\"),")
```
ssmifi, line 47: _Suggestion_: Could also use `strconv.Quote` here, but it is more a consistency thing here and not a real issue.

lexer/doc.go
```diff
@@ -25,6 +25,10 @@
 //
 // The generated `NewLexer` function returns a [DefaultLexer] constructed via
 // [NewDefaultLexer] with every keyword and token type for that language.
+// Grammars with multiple configured languages use [NewMultiLanguageLexer]
+// instead, passing one token type list per language; at lex time the document
+// is routed to its language's token set via [core.LanguageSelector], mirroring
+// the generated parser's entry rule dispatch.
 // [typefox.dev/fastbelt/workspace.DefaultDocumentParser] obtains a [Lexer] from
```
ssmifi: _Minor_: `DefaultDocumentParser` does not exist anymore. Same for the `LexerResult` struct.

lexer/lexer.go
```diff
@@ -113,37 +101,49 @@ func (l *DefaultLexer) Lex(input string) *LexerResult {
...
 		}
+		tokenMaps[li] = tokenMap
+		avgRatios[li] = parallel.NewRunningAverage(defaultTokenRatio)
 	}
 
 	return &DefaultLexer{
-		tokenTypes: tokenTypes,
-		tokenMap:   tokenMap,
-		avgRatio:   parallel.NewRunningAverage(defaultTokenRatio),
+		tokenTypes: languages,
```
ssmifi, line 144: _Question_: Is the `tokenTypes` field actually used anywhere? Where should it be used?

parser/parser.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // Parser defines the interface for parsing tokens (lexer output) into AST nodes.
 type Parser interface {
-	Parse(document *core.Document) *ParseResult
+	Parse(document *core.Document)
 }
 
 type ParseResult struct {
```
ssmifi: This is not used anylonger, right?

test/fixture.go
```diff
@@ -118,15 +118,10 @@ func (f *Fixture) WithMarking(m *TestMarking) *Fixture {
 }
 
 // Parse builds a single in-memory document from content. Markers are extracted before
-// parsing. The URI is "inmemory://test<ext>" where <ext> is the first entry in
-// WorkspaceSrv.FileExtensions (e.g. ".statemachine"), or "inmemory://test" if none
-// are configured.
+// parsing. The URI is "inmemory://test". Use [ParseURI] to specify a different URI.
 func (f *Fixture) Parse(content string) *Doc {
 	f.t.Helper()
 	uri := "inmemory://test"
```
ssmifi: _Question_: Does this actually populate the uri path? The glob matching in the document selector requires a uri path, right?

textdoc/file.go
```diff
@@ -25,21 +24,15 @@ type File struct {
 
 // NewFile creates a new text document file.
 // The content parameter is accepted as a string and stored internally as bytes.
-func NewFile(uri lsp.DocumentURI, languageID string, version int32, content string) (*File, error) {
-	if uri == "" {
```
ssmifi, line 29: _Question_: Why does this need to go?

util/glob/glob.go
```diff
@@ -0,0 +1,70 @@
...
+import "strings"
+
+// Match reports whether path matches a glob pattern. Paths are matched
+// segment-by-segment on '/'. Within a segment, '*' matches any run of
+// non-separator characters and '?' matches a single one. The '**' segment
+// matches zero or more whole segments (crossing separators).
+//
+// ponytail: minimal glob covering the "**/*.ext" defaults + simple hand-written
+// patterns; swap for a glob lib if callers ever need {a,b} / [..] classes.
+func Match(pattern, path string) bool {
```
ssmifi, line 17: _Question_: Is case-sensitivity a concern here? Could it produce headaches on case-insensitive file systems?

util/glob/glob.go
```diff
@@ -0,0 +1,70 @@
...
+package glob
+
+import "strings"
+
+// Match reports whether path matches a glob pattern. Paths are matched
+// segment-by-segment on '/'. Within a segment, '*' matches any run of
+// non-separator characters and '?' matches a single one. The '**' segment
+// matches zero or more whole segments (crossing separators).
+//
+// ponytail: minimal glob covering the "**/*.ext" defaults + simple hand-written
```
ssmifi: "Ponytail"? :)

workspace/builder.go
```diff
@@ -173,11 +185,13 @@ func (s *DefaultBuilder) Build(ctx context.Context, docs []*core.Document, downg
 
 func (s *DefaultBuilder) Reset(doc *core.Document, state core.DocumentState) {
 	switch {
+	case !state.Has(core.DocStateLexed):
+		doc.Tokens = core.TokenSlice{}
+		doc.LexerErrors = []*core.LexerError{}
+		fallthrough
 	case !state.Has(core.DocStateParsed):
```
ssmifi, line 192: _Question_: How does this work now with the `keepState` in the updater? Should `DocStateLexed` be set there as well?

workspace/initializer.go
```diff
@@ -58,21 +57,23 @@ type DefaultIncludeFilter struct {
 
 // Include returns true if the given file system entry should be included.
 // See [DefaultIncludeFilter] for more info.
-func (s *DefaultIncludeFilter) Include(entry DirEntry) bool {
+func (s *DefaultIncludeFilter) Include(entry DirEntry) (bool, string) {
```
ssmifi, line 62: _Suggestion_: Clean up doc in the whole document. There are several stale docs in this file referring to [FileExtension] and [workspace.LanguageID]. Also, here at the Include, the string should be explained.

document.go
```diff
@@ -99,33 +99,32 @@ func NewDocument(textDoc textdoc.Handle) *Document {
...
 }
 
 // DocumentState is a bitmask capturing the already completed build phases of a document.
 type DocumentState uint32
 
 const (
-	// DocStateParsed marks that lexing and parsing completed and produced AST and token data.
-	DocStateParsed DocumentState = 1 << iota // 0x0001
+	// DocStateLexed marks that lexing completed and produced token data.
+	DocStateLexed DocumentState = 1 << iota // 0x0001
```
ssmifi, line 113: `DocStateLexed` is missing from `docStateComplete` and `String()`?

language_selector.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"typefox.dev/fastbelt/textdoc"
+	"typefox.dev/fastbelt/util/glob"
+	"typefox.dev/fastbelt/util/service"
+)
+
+type DocumentMatcher func(uri URI) bool
+
+// DocumentFilter matches a document by LSP language id and/or a glob over the
```
ssmifi: `DocumentSelector`?

language_selector.go
```diff
@@ -0,0 +1,90 @@
...
+
+package fastbelt
+
+import (
+	"typefox.dev/fastbelt/textdoc"
+	"typefox.dev/fastbelt/util/glob"
+	"typefox.dev/fastbelt/util/service"
+)
+
+type DocumentMatcher func(uri URI) bool
```
ssmifi, line 15: As this is an exported symbol, it should probably get a doc, even if self-explanatory.

ssmifi (commented):
Looks solid, thanks. I have a couple of questions to get a better understanding before merging and some documentation seems to be stale.

cmd/fastbelt/generate.go
msujew: We still have the open issue https://github.com/TypeFox/fastbelt/issues/83, which I wouldn't try to tackle in this PR.

cmd/build.go
msujew: Oops, outdated docs. I'll fix that up.

cmd/build.go
msujew: We discussed this internally, and decided to just go without an import mechanism for now. Meaning no exploration of files beyond the current directory.

cmd/build.go
msujew: I plan to implement something for this before the 0.2 release.

examples/multilang/completion_parser_gen.go
msujew: Oops, forgot to update after making some changes in `parser_generator`.

internal/generator/lexer_generator.go
msujew: Yes, that will come with #127.

util/glob/glob.go
msujew: Claude doing claude things...

test/fixture.go
msujew: It doesn't actually populate it, but the glob matching will generally fail anyway, because it doesn't have a file extension. This is fine for projects with just a single language.

textdoc/file.go
msujew: There was no real reason why you couldn't construct a document with an empty URI or language ID. It only makes the API more complex.

util/glob/glob.go
msujew: Maybe, but we don't bother in Langium as well with that, and no one complained so far.

workspace/builder.go
msujew: Good catch!

examples/arithmetics/services.go
```diff
@@ -13,8 +14,12 @@ import (
 
 // SetupServices sets up the base services for the arithmetics language.
 func SetupServices(sc *service.Container) {
-	service.Put[workspace.LanguageID](sc, "arithmetics")
-	service.Put[workspace.FileExtensions](sc, []string{".arithmetics"})
+	service.Put[core.LanguageSelector](
+		sc,
+		core.NewDefaultLanguageSelector(sc,
+			core.NewDocumentSelectorWithPatterns("arithmetics", "**/*.arithmetics"),
```
Lotes, line 20: Puh, a document selector can also influence the folders up the file hierarchy. Also here you seem to be forced to use „\*\*/“… this is broader than an array of file extensions. Not sure if we should go for this. You always need to prepend „\*\*/*“ to your „.ext“.
Is this case-insensitive btw?

msujew: > Also here you seem to be forced to use „**/“… this is broader than an array of file extensions.

You're not forced to do it. Adopters can implement their own `FileExtensionDocumentSelector` if they want to. If absolutely required, we can also implement this in the framework, but I'm not really convinced that this is better than the glob pattern.

I'm not happy how this works in Langium, because the idea to use file extensions as the defining identifying feature of languages has a large negative impact on the API design.

> Is this case-insensitive btw?

No, but neither is Langium's file extension comparison AFAIK.

Lotes: OK, I am not really for or against it. It just feels unusual.
But it serves the same purpose as before and goes even beyond that, which might be a good sign to keep it.

cmd/build.go
```diff
@@ -0,0 +1,333 @@
...
+	"typefox.dev/fastbelt/workspace"
+	"typefox.dev/lsp"
+)
+
+// Language configures one language served by the generated language server: the
+// grammar entry rule it parses from and the [Selector] that claims its
+// documents.
+type Language struct {
+	Entry      string
+	LanguageID string
```
Lotes, line 33: Add a lexer mode, after #127 is merged.

internal/languages/multilang/multilang.fb
```diff
@@ -0,0 +1,15 @@
+grammar MultilangModel
```
Lotes, line 1: I thought this PR is also about multi file support.
But yes, this might be a smaller step towards this direction.

document.go
```diff
@@ -99,38 +99,40 @@ func NewDocument(textDoc textdoc.Handle) *Document {
...
+	DocStateExportedSymbols // 0x0004
 	// DocStateImportedSymbols marks that symbols from other documents were imported.
-	DocStateImportedSymbols // 0x0004
+	DocStateImportedSymbols // 0x0008
 	// DocStateLocalSymbols marks that local symbols for this document were collected.
-	DocStateLocalSymbols // 0x0008
+	DocStateLocalSymbols // 0x0010
 	// DocStateLinked marks that cross-references were linked.
-	DocStateLinked // 0x0010
+	DocStateLinked // 0x0020
```
Lotes, line 123: Why the comment "//0x0001" ...? Who needs it?
The only reason in my eyes: to not forget the flag nature of this constant group.

Alternative: Rename the constants to reflect the flag nature.

* DocStateFlagLinked
* DocStateBitMaskLinked

cmd/build.go
Lotes: Or combine it within the syntax of the entry rule:

```
entry XRule on mode YMode: ...
```

Omitting `on mode YMode` falls back to the `default` mode.

cmd/build.go
sailingKieler: Still outdated docs, right?

language_selector.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	if d.Matcher != nil && d.Matcher(uri) {
+		return true
+	}
+	return false
+}
+
+// LanguageSelector resolves a document URI to the index of the owning language
+// (into the configured languages), or -1 if none match.
+type LanguageSelector interface {
```
sailingKieler, line 59: On the first impression this felt a bit over engineered to me.
After thinking about it for a moment or tow, I came to the conclusion that this mainly bridges the lack of multi-bindings in our dependency container, here, multiple `DocumentSelectors`, correct?

I would appreciate a corresponding hint on that here in the docs.

internal/languages/multilang/gen/main.go
```diff
@@ -0,0 +1,34 @@
...
+				Entry:      "Farewell",
+				LanguageID: "farewell",
+				Patterns:   []string{"**/*.bye"},
+			},
+		},
+		Verbose: true,
+	}
+	if err := ctx.Build(); err != nil {
+		log.Fatalf("multilang generation failed: %v", err)
+	}
```
sailingKieler, line 33: Is this route really needed for the multi lange setup?

Why not having it configured manually in the scaffold provided `services.go`, as you did in this language impl? Coding agents will be able to do that shortly.

Hence, I'm favor of skipping the generation in `service_generator.go`.

cmd/fastbelt/generate.go
sailingKieler: +1

internal/generator/services_generator.go
```diff
@@ -26,6 +33,26 @@ func GenerateServices(grammr grammar.Grammar, packageName string) string {
...
+						}
+					})
+					n.AppendLine("),")
+				})
+				n.AppendLine(")")
+			})
+			n.AppendLine("}")
+		}
```
sailingKieler, line 55: I don't see the need/benefit of this, see also my remark in `languages/multilang/gen/main.go`.

Having it codified in `services.go` based on an initial template should be fine IMHO.
The single rule validation in `cmd/fastbelt/generator.go` could be suppressed by a cmd line switch that is to be added in the `// go:generate ...` config.

cmd/build.go
sailingKieler: Agree with @Lotes.

However, this idea mixes concepts that don't belong together:
> ```
> entry XRule on mode YMode: ...
> ```

I don't support this proposal.

internal/generator/lexer_generator.go
```diff
@@ -142,27 +144,131 @@ func GenerateLexer(grammr grammar.Grammar, packageName string, tokenTypes Genera
...
+							nnn.AppendLine(name, ",")
+						}
+					}
+					for _, token := range tokens {
+						if name := GeneratedTokenName(token); reachable[name] {
+							nnn.AppendLine(name, ",")
+						}
+					}
+				})
+				nn.AppendLine("},")
```
sailingKieler, line 192: uuuhhh this is sophisticated.

This should be documented very prominently, specifically the fact parser (entry) rule names/ids are used in the lexer setup.

lexer/doc.go
```diff
@@ -25,10 +25,10 @@
...
 // The generated `NewLexer` function returns a [DefaultLexer] constructed via
 // [NewDefaultLexer] with every keyword and token type for that language.
-// [typefox.dev/fastbelt/workspace.DefaultDocumentParser] obtains a [Lexer] from
-// the service container, calls [Lexer.Lex], and stores [LexerResult.Tokens],
-// [LexerResult.Comments], and [LexerResult.Errors] on the document before
-// parsing.
+// Grammars with multiple configured languages use [NewMultiLanguageLexer]
+// instead, passing one token type list per language; at lex time the document
+// is routed to its language's token set via [core.LanguageSelector], mirroring
+// the generated parser's entry rule dispatch.
```
sailingKieler, line 31: I `lexer_gen.go` I asked for prominent documentation of this fact.
Okay, it is already stated here but I didn't get the consequences of this fairly compact statement.

I strongly advocate to extend it, explicit state that reachable token sets from each entry rules are determined.
That this way ambiguities among the entire token definitions set are tolerated as long the sets reachable from the entry rules a unambiguous (enough).
And that this is also beneficial performance wise (less dispatches)

internal/languages/multilang/multilang.go
sailingKieler, line 1: Let's adhere the common pattern of naming this file `doc.go`.

```
//go:generate go run ./gen
```
can go to `services.go`, like in the other languages.

sailingKieler (commented):
Thanks @msujew for this effort! Great step forward.

Actually, I doubt the necessity and benefit for the additional language build customization for the multi language aspect. Other concerns might justify that.
I left some comments inline.

The other aspect is the documentation of impact of multiple entry rules on the lexer.
That is something I didn't expect in the first place, and others (newbies) probably also don't. See also below.

cmd/build.go
msujew: Yes, I would keep the entry lexer mode local to the `Language` struct value.

PR #131, opened 2026-07-24
Add per-language lexers to eliminate cross-language keyword pollution

cdietrich:
When a grammar has multiple entry rules (multi-language mode), each entry rule now gets its own NewLexerFor\<i\>() factory that includes only the keywords reachable from that entry point via BFS over rule calls.

A generated documentParser struct (language-aware) selects the right per-language lexer based on LanguageSelector.Select(uri) before tokenising, preventing keywords from one sub-language shadowing identifiers in another.

Changes:
- internal/generator/reachability.go: KeywordsReachableFrom BFS traversal
- internal/generator/lexer_generator.go: GenerateLexer emits NewLexerFor\<i\>()
- internal/generator/services_generator.go: GenerateServices emits language-aware DocumentParser and uses service.Override to replace the workspace default
- cmd/build.go: pass entryRules to GenerateLexer and GenerateServices

Claude-Session: https://claude.ai/code/session_01R7GR5jLbxDCbKdhASefCMH


msujew:
Thanks, this has been superseded by a change in https://github.com/TypeFox/fastbelt/pull/130, which implements this as well.

PR #132, opened 2026-07-24
Fix order of workspace builder reset

msujew:
Fixes a minor issue introduced in #126. Adds a test that currently fails on `main` for regression testing/validation.

### DIFF
--- a/workspace/builder.go
+++ b/workspace/builder.go
@@ -183,11 +183,11 @@ func (s *DefaultBuilder) Reset(doc *core.Document, state core.DocumentState) {
 	case !state.Has(core.DocStateExportedSymbols):
 		doc.ExportedSymbols = nil
 		fallthrough
-	case !state.Has(core.DocStateImportedSymbols):
-		doc.ImportedSymbols = nil
-		fallthrough
 	case !state.Has(core.DocStateLocalSymbols):
 		doc.LocalSymbols = nil
+		fallthrough
+	case !state.Has(core.DocStateImportedSymbols):
+		doc.ImportedSymbols = nil
 		fallthrough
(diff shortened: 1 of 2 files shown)


workspace/builder.go
```diff
@@ -183,12 +183,12 @@ func (s *DefaultBuilder) Reset(doc *core.Document, state core.DocumentState) {
...
 		fallthrough
-	case !state.Has(core.DocStateImportedSymbols):
-		doc.ImportedSymbols = nil
-		fallthrough
 	case !state.Has(core.DocStateLocalSymbols):
 		doc.LocalSymbols = nil
 		fallthrough
+	case !state.Has(core.DocStateImportedSymbols):
+		doc.ImportedSymbols = nil
+		fallthrough
```
Lotes, line 191: Is it true that a fallthrough ignores the next cases condition?

msujew: No, the opposite. `fallthrough` will fall through to the next case. The semantics between Go and C are switched. In C, `fallthrough` is the default, whereas `break` needs to be used to escape the switch. In Go `break` is the default, and `fallthrough` needs to be used to continue with the next case.

ssmifi: That is actually a smart change hopefully satisfying both camps. :-)

ssmifi (approved):
LGTM

sailingKieler (commented):
Had to think about it for a moment.

Makes sense to me, as `ImportedSymbols` is probably more volatile in practice as it's to be updated during incremental builds triggered by changes in other files — although there's no conceptual dependency between `LocalSymbols` and `ImportedSymbols`.

PR #133, opened 2026-07-24
Reduce AST memory allocations

msujew:
The parser currently allocates a lot of empty arrays that aren't really necessary. For most purposes, `nil` and `[]` (empty) slices are essentially the same. This change removes all of the array initialization code, which reduces retained memory, reduces the amount of intermediate allocated memory, and improves parsing performance.

Adds a benchmark for the arithmetics language, which is affected by this change more than the statemachine language.

### DIFF
--- a/ast.go
+++ b/ast.go
@@ -39,9 +39,9 @@ type AstNode interface {
 	SetContainer(container AstNode, containerField unique.Handle[string], index int)
 	// Tokens returns the tokens associated with the node.
 	Tokens() []*Token
-	// SetToken appends token to the node's token list.
-	SetToken(token *Token)
-	// SetTokens replaces the node's token list with tokens.
+	// AppendToken appends token to the node's token list.
+	AppendToken(token *Token)
+	// SetTokens replaces the node's token list with another list of tokens.
 	SetTokens(tokens []*Token)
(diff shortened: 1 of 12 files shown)


ast.go
```diff
@@ -181,7 +181,14 @@ func (node *AstNodeBase) TextRange() TextRange {
 // SetToken appends token to the node's token list.
 func (node *AstNodeBase) SetToken(token *Token) {
```
Lotes: SET „appends“ to the list… why not ADD or APPEND?
A confusing convention?

ast.go
```diff
@@ -449,31 +446,28 @@ type CompositeNode interface {
...
 
 // Owner returns the AST node that owns this string unit.
-func (node *CompositeNodeBase) Owner() AstNode {
+func (node *compositeNode) Owner() AstNode {
 	return node.container
 }
 
-// String returns the concatenated token images of node, caching the computed value.
-func (node *CompositeNodeBase) String() string {
+func (node *compositeNode) String() string {
```
ssmifi, line 470: _Cosmetic_: `CompositeNodeBase` is still mentioned in the comment of `ast_test.go`.

ssmifi (approved):
LGTM

PR #134, opened 2026-07-27
Support infix rules

### LINKED ISSUE #115: Arithmetics example: Exponentiation not right-associative yet
Current implementation:

```
Exponentiation returns Expression:
    Modulo ({BinaryExpression.Left=current} Operator="^" Right=Modulo)*;
```

In [Langium](https://langium.org/docs/reference/grammar-language/infix-operators/manual-implementation/)/Xtext, this would be actually:

```
Exponentiation returns Expression:
    Modulo ({BinaryExpression.Left=current} Operator='^' Right=Exponentiation)?;
```

msujew:
Closes https://github.com/TypeFox/fastbelt/issues/115 (as a side effect).

Adds infix rule support, similar to how Langium did it (but a bit simpler, since we don't need CST support). 

Implements the support by rewriting infix rules into parser rules of the shape `Expr (op Expr)*`.

### DIFF
--- a/cmd/fastbelt/generate.go
+++ b/cmd/fastbelt/generate.go
@@ -14,7 +14,7 @@ import (
 
 	core "typefox.dev/fastbelt"
 	"typefox.dev/fastbelt/internal/generator"
-	grammarPkg "typefox.dev/fastbelt/internal/grammar"
+	"typefox.dev/fastbelt/internal/grammar"
 	"typefox.dev/fastbelt/textdoc"
 	"typefox.dev/fastbelt/util/service"
 	"typefox.dev/fastbelt/workspace"
@@ -54,7 +54,7 @@ func runGenerateCLI(opts generateOptions) error {
 		return err
 	}
 
-	sc := grammarPkg.CreateServices()
+	sc := grammar.CreateServices()
 	file, _ := textdoc.NewFile(lsp.URIFromPath(grammarPath), "fb", 0, string(grammarText))
 
 	document := core.NewDocument(file)
@@ -97,17 +97,17 @@ func runGenerateCLI(opts generateOptions) error {
 		return fmt.Errorf("aborting code generation due to %d errors", errCount)
 	}
 
-	grammar, ok := document.Root.(grammarPkg.Grammar)
+	gram, ok := document.Root.(grammar.Grammar)
 	if !ok {
 		return fmt.Errorf("parser result is not a Grammar")
 	}
-	entryRule, err := validateEntryRule(grammar)
+	entryRule, err := validateEntryRule(gram)
 	if err != nil {
 		return err
 	}
 	// Desugar infix rules once, so every generator below sees the synthesized
 	// operator token groups and flat rule bodies.
-	if err := grammarPkg.ExpandInfixRules(grammar); err != nil {
+	if err := grammar.ExpandInfixRules(gram); err != nil {
 		return err
 	}
 
@@ -122,46 +122,46 @@ func runGenerateCLI(opts generateOptions) error {
 	}
 
 	if err := writeFile("linker", filepath.Join(outputPath, "linker_gen.go"),
-		generator.GenerateLinker(grammar, packageName)); err != nil {
+		generator.GenerateLinker(gram, packageName)); err != nil {
 		return err
 	}
 	if err := writeFile("types", filepath.Join(outputPath, "types_gen.go"),
-		generator.GenerateTypes(grammar, packageName)); err != nil {
+		generator.GenerateTypes(gram, packageName)); err != nil {
 		return err
 	}
-	tokenTypes := generator.GenerateTokenTypes(grammar)
-	atnData := generator.BuildParserATNData(grammar, tokenTypes)
+	tokenTypes := generator.GenerateTokenTypes(gram)
+	atnData := generator.BuildParserATNData(gram, tokenTypes)
 	if err := writeFile("parser", filepath.Join(outputPath, "parser_gen.go"),
-		generator.GenerateParser(grammar, entryRule, packageName, tokenTypes, atnData)); err != nil {
+		generator.GenerateParser(gram, entryRule, packageName, tokenTypes, atnData)); err != nil {
 		return err
 	}
 	if err := writeFile("completion-parser", filepath.Join(outputPath, "completion_parser_gen.go"),
-		generator.GenerateCompletionParser(grammar, entryRule, packageName, tokenTypes, atnData)); err != nil {
+		generator.GenerateCompletionParser(gram, entryRule, packageName, tokenTypes, atnData)); err != nil {
 		return err
 	}
 	if err := writeFile("parser-lookahead", filepath.Join(outputPath, "parser_lookahead_gen.go"),
-		generator.GenerateParserLookahead(grammar, packageName, tokenTypes, atnData)); err != nil {
+		generator.GenerateParserLookahead(gram, packageName, tokenTypes, atnData)); err != nil {
 		return err
 	}
 	if err := writeFile("completion", filepath.Join(outputPath, "completion_gen.go"),
-		generator.GenerateCompletion(grammar, packageName)); err != nil {
+		generator.GenerateCompletion(gram, packageName)); err != nil {
 		return err
 	}
 	if err := writeFile("lexer", filepath.Join(outputPath, "lexer_gen.go"),
-		generator.GenerateLexer(grammar, packageName, tokenTypes)); err != nil {
+		generator.GenerateLexer(gram, packageName, tokenTypes)); err != nil {
 		return err
 	}
 	if err := writeFile("services", filepath.Join(outputPath, "services_gen.go"),
-		generator.GenerateServices(grammar, packageName)); err != nil {
+		generator.GenerateServices(gram, packageName)); err != nil {
 		return err
 	}
 	if err := writeFile("atn", filepath.Join(outputPath, "atn_gen.go"),
-		generator.GenerateATN(grammar, packageName, tokenTypes)); err != nil {
+		generator.GenerateATN(gram, packageName, tokenTypes)); err != nil {
 		return err
 	}
 	if opts.atn {
 		if err := writeFile("atn-md", filepath.Join(outputPath, "atn.md"),
-			generator.GenerateATNMarkdown(grammar, packageName, tokenTypes)); err != nil {
+			generator.GenerateATNMarkdown(gram, packageName, tokenTypes)); err != nil {
 			return err
 		}
 	}
@@ -169,8 +169,8 @@ func runGenerateCLI(opts generateOptions) error {
 	return nil
 }
 
-func validateEntryRule(g grammarPkg.Grammar) (grammarPkg.ParserRule, error) {
-	var entries []grammarPkg.ParserRule
+func validateEntryRule(g grammar.Grammar) (grammar.ParserRule, error) {
+	var entries []grammar.ParserRule
 	for _, rule := range g.Rules() {
 		if rule.IsEntry() {
 			entries = append(entries, rule)
--- a/cmd/fastbelt/generate.go
+++ b/cmd/fastbelt/generate.go
@@ -14,7 +14,7 @@ import (
 
 	core "typefox.dev/fastbelt"
 	"typefox.dev/fastbelt/internal/generator"
-	"typefox.dev/fastbelt/internal/grammar"
+	grammarPkg "typefox.dev/fastbelt/internal/grammar"
 	"typefox.dev/fastbelt/textdoc"
 	"typefox.dev/fastbelt/util/service"
 	"typefox.dev/fastbelt/workspace"
@@ -54,7 +54,7 @@ func runGenerateCLI(opts generateOptions) error {
 		return err
 	}
 
-	sc := grammar.CreateServices()
+	sc := grammarPkg.CreateServices()
 	file, _ := textdoc.NewFile(lsp.URIFromPath(grammarPath), "fb", 0, string(grammarText))
 
 	document := core.NewDocument(file)
@@ -97,12 +97,17 @@ func runGenerateCLI(opts generateOptions) error {
 		return fmt.Errorf("aborting code generation due to %d errors", errCount)
 	}
 
-	grammar, ok := document.Root.(grammar.Grammar)
+	grammar, ok := document.Root.(grammarPkg.Grammar)
 	if !ok {
 		return fmt.Errorf("parser result is not a Grammar")
 	}
 	entryRule, err := validateEntryRule(grammar)
 	if err != nil {
+		return err
+	}
+	// Desugar infix rules once, so every generator below sees the synthesized
+	// operator token groups and flat rule bodies.
+	if err := grammarPkg.ExpandInfixRules(grammar); err != nil {
 		return err
 	}
 
@@ -164,8 +169,8 @@ func runGenerateCLI(opts generateOptions) error {
 	return nil
 }
 
-func validateEntryRule(g grammar.Grammar) (grammar.ParserRule, error) {
-	var entries []grammar.ParserRule
+func validateEntryRule(g grammarPkg.Grammar) (grammarPkg.ParserRule, error) {
+	var entries []grammarPkg.ParserRule
 	for _, rule := range g.Rules() {
 		if rule.IsEntry() {
 			entries = append(entries, rule)
--- a/cmd/fastbelt/generate.go
+++ b/cmd/fastbelt/generate.go
@@ -14,7 +14,7 @@ import (
 
 	core "typefox.dev/fastbelt"
 	"typefox.dev/fastbelt/internal/generator"
-	"typefox.dev/fastbelt/internal/grammar"
+	grammarPkg "typefox.dev/fastbelt/internal/grammar"
 	"typefox.dev/fastbelt/textdoc"
 	"typefox.dev/fastbelt/util/service"
 	"typefox.dev/fastbelt/workspace"
@@ -54,7 +54,7 @@ func runGenerateCLI(opts generateOptions) error {
 		return err
 	}
 
-	sc := grammar.CreateServices()
+	sc := grammarPkg.CreateServices()
 	file, _ := textdoc.NewFile(lsp.URIFromPath(grammarPath), "fb", 0, string(grammarText))
 
 	document := core.NewDocument(file)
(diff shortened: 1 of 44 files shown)


examples/arithmetics/arithmetics.fb
```diff
@@ -58,19 +58,24 @@ Evaluation:
...
+    BinaryExpression
+
+infix BinaryExpression on PrimaryExpression:
+    "%"
+    > right assoc "^"
+    > Multiplication
+    > Addition
+
+// Tests that Fastbelt can handle operators nested in token groups
+token group Multiplication {
```
Lotes: Suggestion: `MultiplicationOp` and `AdditionOp`

examples/arithmetics/infix_test.go
```diff
@@ -0,0 +1,150 @@
...
+func TestExponentiationRightAssociative(t *testing.T) {
+	expr := parseExpression(t, "2 ^ 3 ^ 2 ^ 1")
+	assert.Equal(t, "(2 ^ (3 ^ (2 ^ 1)))", printExpression(expr))
+
+	expr = parseExpression(t, "1 - 2 - 3")
+	assert.Equal(t, "((1 - 2) - 3)", printExpression(expr))
+
+	// Mixed with tighter ("%") and looser ("*") levels.
+	expr = parseExpression(t, "1 * 2 ^ 3 % 4 ^ 5")
+	assert.Equal(t, "(1 * (2 ^ ((3 % 4) ^ 5)))", printExpression(expr))
```
Lotes, line 72: This line reads weird, until you mind the comment and the special grammar with % as strongest operator.

grammar/doc.go
```diff
@@ -358,6 +358,55 @@
...
+//
+// The rule name must resolve to an interface declaring the fields Left,
+// Right (of an interface type both the rule's return type and node type are
+// assignable to), and Operator (of type string). The rule after "on" is the
+// operand rule, parsed between operators. Precedence groups are separated by
+// ">" and ordered tightest-binding first: in the example, "%" binds tightest
+// and "+" | "-" bind loosest, so "1 + 2 * 3" parses as "1 + (2 * 3)".
+//
+// Groups are left-associative by default; prefix a group with "right assoc"
+// (or explicitly "left assoc") to control associativity:
```
Lotes: We could also think of a "non assoc" for operators like less and greater. But I would prefer a extra validation I think

internal/grammar/grammar.fb
```diff
@@ -171,6 +173,26 @@ CompositeGroup returns Element: CompositeElement ({Group.Elements+=current} Elem
...
+InfixRule: "infix" Name=ID "on" Call=RuleCall ("returns" ReturnType=[Interface:ID])? ":"
+    Groups+=PrecedenceGroup (">" Groups+=PrecedenceGroup)* ";"?
+
+interface PrecedenceGroup {
+    Associativity string
+    Operators     []Assignable
+}
+
+PrecedenceGroup:
+    (Associativity=("left" | "right") "assoc")?
```
ssmifi: _Discussion_: I am wondering if we need the `assoc` keyword at all.

ssmifi: For me it is more or less noise. `left` and `right` suffice and it is not even a full word. :) 
I know that this is the Langium syntax, so maybe just make it optional.

internal/grammar/validator.go
```diff
@@ -711,6 +727,32 @@ func isAssignableTo(ctx context.Context, source Assignable, fieldType FieldType,
 					core.WithCode(ValidateAssignmentType),
 				))
 			}
+		case InfixRule:
```
ssmifi, line 703: This is identical to `ParserRule` with the exception of one line. Can we combine this into `case ParserRule, InfixRule:` and just differ in `FindReturnType`?

internal/generator/parser_generator.go
```diff
@@ -1097,6 +1111,15 @@ func generateRuleCallParser(node codegen.Node, context *ParserGeneratorContext,
 				n.AppendLine("result ", eq, " p.Parse", t.Name(), "()")
 			}
 			n.AppendLine("p.state.ExitRule()")
+		case grammar.InfixRule:
```
ssmifi, line 1293: Could use just `case grammar.ParserRule, grammar.InfixRule:`

internal/generator/infix_generator.go
```diff
@@ -0,0 +1,183 @@
...
+			n.AppendLine(syncCall)
+		}
+		n.AppendLine("for ", guardCall(context, inner), " {")
+		n.Indent(func(in codegen.Node) {
+			in.AppendLine("if token := p.state.Consume(", GeneratedTokenName(operatorGroup), "); token != nil {")
+			in.Indent(func(iin codegen.Node) {
+				iin.AppendLine("operators = append(operators, token)")
+			})
+			in.AppendLine("}")
+			emitOperand(in, secondCall)
```
ssmifi, line 92: _Question_: Is it safe to emit the operand, even if the operator wasn't appended before?

parser/infix.go
```diff
@@ -0,0 +1,69 @@
...
+	// rec builds the tree for operators[lo:hi] and their surrounding parts.
+	var rec func(lo, hi int) T
+	rec = func(lo, hi int) T {
+		if lo == hi {
+			if lo < len(parts) {
+				return parts[lo]
+			}
+			return zero
+		}
+		pivot := lo
```
ssmifi: _Question_: Can this degrade to O(n²), e.g., at long same precedence chains? 
Probably not overly relevant in praxis, I was just wondering. :)

cmd/fastbelt/generate.go
```diff
@@ -14,7 +14,7 @@ import (
 
 	core "typefox.dev/fastbelt"
 	"typefox.dev/fastbelt/internal/generator"
-	"typefox.dev/fastbelt/internal/grammar"
+	grammarPkg "typefox.dev/fastbelt/internal/grammar"
```
ssmifi: _Consideration_: Wouldn't it be more go-idiomatic to just use `g` as the local variable name and keep the `grammar` import?

ssmifi (commented):
Looks well-designed and -tested. There are one design and some minor questions left.

parser/infix.go
msujew: Yes, that's the worse case for this behavior. It's probably nothing to worry about, as even then, it is likely much faster than the alternative (which is nested rule calls).

msujew: Nevermind, I implemented a linear algorithm, it's much faster :)

internal/generator/infix_generator.go
msujew: Emitting the operand is definitely safe. The question should be: Is it safe to append the operand if the operator didn't exist ;)

In that case, it's still safe, since we cannot enter the `for` loop that is emitted before that. The lookahead done there asserts that the next token is the expected operator token. And even then, error recovery should ensure that this works as expected.

grammar/doc.go
```diff
@@ -358,6 +358,55 @@
 // becomes the current object. The operator += is also valid for tree-rewriting
 // actions on slice properties.
 //
+// # Infix Rules
```
Lotes, line 361: We could also explain

* what the difference between associativity and precedence is
* how it is done without infix notation, and that infix is faster than manual
* that this is useful to reduce ambiguities in expressions
  * and that reducing ambiguities, is increasing parsing performance

internal/grammar/infix.go
```diff
@@ -0,0 +1,200 @@
...
+		groupName := InfixOperatorGroupName(rule)
+		if findRuleByName(g, groupName) != nil {
+			return fmt.Errorf("cannot generate operator token group for infix rule '%s': the name '%s' is already taken", rule.Name(), groupName)
+		}
+		operand := resolveOperandRule(rule, ctx)
+		if operand == nil {
+			return fmt.Errorf("could not resolve the operand rule of infix rule '%s'", rule.Name())
+		}
+
+		tokenGroup, err := synthesizeOperatorGroup(g, rule, groupName, ctx)
```
Lotes, line 60: What if an operator is not a terminal? Looking at the grammar, this should be an option.

internal/grammar/infix.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+	}
+	tokenGroup.SetContainer(g, fieldNameTokenGroups, len(g.TokenGroups()))
+	g.SetTokenGroupsItem(tokenGroup)
+	return tokenGroup, nil
+}
+
+// synthesizeInfixBody attaches the flat "operand (operator operand)*" body to
+// the infix rule.
+func synthesizeInfixBody(rule InfixRule, operand ParserRule, operatorGroup TokenGroup) {
```
Lotes, line 147: I feel a bit lost here.

1. `operand (operator operand)*` only handles left association. What about right association? Or what do I read wrong?
2. Is this about creating a AST for the infix operators?

parser/infix_test.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return &infixTestNode{left: left, right: right, op: op}
+}
+
+func TestBuildInfixTree(t *testing.T) {
+	typeIds := map[string]int{"%": 1, "^": 2, "*": 3, "/": 4, "+": 5, "-": 6, "=": 7}
+	// Mirrors the arithmetics grammar: "%" > "^" > "*" | "/" > "+" | "-",
+	// plus a loosest right-associative "=" level.
+	precedence := map[int]InfixPrecedence{
+		typeIds["%"]: {Level: 0},
```
Lotes, line 58: Just because it reads weird:
Instead of the modulo operator can we instead use the [tetration](https://en.wikipedia.org/wiki/Tetration) operator?
Either "!!" or "↑↑" (decimal: 8593, hex: 2191, entity: \&uarr).

internal/grammar/infix_validator_test.go
```diff
@@ -0,0 +1,144 @@
...
+func TestInfixRuleOperatorGroupNameTaken(t *testing.T) {
+	f := test.New(t, CreateServices())
+	doc := f.Parse(`
+		grammar Test;
+	` + infixInterfaces + `
+		token BinaryExpressionOperator: /@+/;
+		infix <|1:BinaryExpression|> on PrimaryExpression: "+"
+	` + commonTokens)
+	doc.ExpectDiagnostic("1").WithSeverity(core.SeverityError).WithCode(ValidateInfixOperatorGroupName)
+}
```
Lotes, line 144: What is the problem here? Is `BinaryExpressionOperator` a reserved name when using infix?

msujew: Yes, the `<infix-rule-name>Operator` is a token group that is automatically generated as an optimization (so lookahead just needs to do a `BinaryExpressionOperator.Matches(p.LA(1))` to figure out whether to continue parsing the binary expression).

internal/grammar/infix.go
msujew: Both associativity directions aren't handled in the parsing phase. They are only relevant for the AST. I.e. whether an expression like `A + B + C` is `(A + B) + C` or `A + (B + C)` is a pure AST mutation. We can apply this as a post-processing on the parsing. Therefore, we first parse all operators and operands into an array, and then apply a transformation that produces the final AST, based on the predetermined precedence.

grammar/doc.go
msujew: How would an AST for a non-associative operation look like? It wouldn't be binary (i.e. just left vs right) anymore.

internal/grammar/infix.go
msujew: We have a validation that prevents this.

grammar/doc.go
msujew: Thanks, I've adjusted the documentation a bit.

> what the difference between associativity and precedence is

They are unrelated. I think this is clearer with the adjusted documentation.

> that this is useful to reduce ambiguities in expressions, and that reducing ambiguities, is increasing parsing performance.

This is not true? It's increasing performance by flattening the parse tree into one single loop, instead of a list of recursive parsing calls. Ambiguity isn't relevant here.

parser/infix_test.go
msujew: What is problematic about the modulo operator? We use the same in the arithmetics language in Langium.

grammar/doc.go
Lotes: It would be still binary. But you could not answer the famous associativity question "which operator is executed first?". You execute only one.

Like `... < a < ...`, the first one is allowed, the second would be a syntax error.

In Xtext this was accomplished by this rule (from [here](https://www.typefox.io/blog/parsing-expressions-with-xtext/)):

```langium
Addition returns Expression:
  Multiplication ({Addition.left=current} '+' right=Multiplication)?;
```

parser/infix_test.go
Lotes: True, it just feels wrong ^^... Modulo has normally a similar level as multiplication and division. Also the tests then felt not "natural" in my eyes.

On the other hand you would have to explain the new operator `!!`.

grammar/doc.go
msujew: This doesn't really mesh well with the `expr (op expr)*` design we're currently using. Also, I'm not sure a syntax error even is a good idea here. I would leave this up for the adopter to identify via a custom validation.

parser/infix_test.go
msujew: I would keep it as is it right now.

grammar/doc.go
Lotes: I was looking at it from a different perspective, like the adopter has no idea what associativity and precedence are...

The comment about performance was meant to show that precedence and associativity are in general are important to reduce ambiguity.

For the sake of simplicity a grammar like:

```antlr
Expression: Expression '+' Expression
          | Expression '-' Expression
          | ...
          ;
```

would be more desirable. Easy to read for humans, but hard to parse for machines, because of ambiguities (which can be resolved with techniques like LL(*), a PEG parser or a backtracking parser(?)).

---

From a perspective "I know already how to make operators work without infix syntax" these comments might be not needed :-/

msujew: > I was looking at it from a different perspective, like the adopter has no idea what associativity and precedence are...

Aren't those basic math terms? I think we can expect some amount of knowledge about these things from our adopters.

> For the sake of simplicity a grammar like:

Ah, I see. I didn't think about people coming from LR parser generators (where this is the norm). I'll include this in the documentation.

cmd/fastbelt/generate.go
sailingKieler, line 1: Why the renaming here? Will cause merge conflict with changes of #130.

internal/grammar/infix_validator_test.go
```diff
@@ -0,0 +1,144 @@
...
+			Left Expression
+			Operator string
+			Right Expression
+		}
+		PrimaryExpression returns Expression: Value=ID
+		infix <|1:BinaryExpression|> on PrimaryExpression returns Other: "+"
+	` + commonTokens)
+	diag := doc.ExpectDiagnostic("1")
+	diag.WithSeverity(core.SeverityError)
+	diag.WithCode(ValidateInfixReturnType)
```
sailingKieler, line 95: is `diag` used anywhere? otherwise, please make consistent with other tests.

grammar/doc.go
```diff
@@ -350,14 +350,77 @@
...
+//
+// Parsing infix expressions (i.e. "a + b * c") is a common problem in language
+// design. Some parser generators solve this by writing a list of alternatives for
+// each precedence level, like this:
+//
+//	Expression: Expression "+" Expression
+//	          | Expression "*" Expression
+//	          | PrimaryExpression
+//
+// In this example, the "*" operator binds tighter than "+", so "a + b * c" parses
```
sailingKieler, line 371: You meant this, right?
```suggestion
// In this example, the "*" operator is supposed to bind tighter than "+", so "a + b * c" parses
```

grammar/doc.go
```diff
@@ -350,14 +350,77 @@
...
+// as "a + (b * c)". This however, only works in parser generators that support left
+// recursion. Since Fastbelt does not support left recursion, the grammar would have
+// to be rewritten to avoid it:
+//
+//	Expression: Addition
+//	Addition: Multiplication ("+" Multiplication)*
+//	Multiplication: PrimaryExpression ("*" PrimaryExpression)*
```
sailingKieler, line 378: Please add a dummy definition of `PrimaryExpression` for completeness.

grammar/doc.go
```diff
@@ -350,14 +350,77 @@
...
+//	Expression: BinaryExpression
+//
+//	infix BinaryExpression on PrimaryExpression:
+//	    "%"
+//	    > right "^"
+//	    > "*" | "/"
+//	    > "+" | "-"
+//
+// The rule name must resolve to an interface declaring the fields Left,
+// Right (of an interface type both the rule's return type and node type are
```
sailingKieler, line 399: What is `node` referring to?
Did you mean this? Then reorder the subsequent sentences, please.
```suggestion
// Right (of an interface type both the rule's return type and operand rule's type are
```

grammar/doc.go
```diff
@@ -350,14 +350,77 @@
...
+//
+// Operators can be keywords, token references, or token group references:
+//
+//	token group MulOp { "*" "/" }
+//	infix BinaryExpression on PrimaryExpression:
+//	    MulOp
+//	    > "+" | "-"
+//
+// The return type of an infix rule defaults to the operand rule's return
+// type; "returns" overrides it. A lone operand is returned unchanged - no
```
sailingKieler, line 418:
```suggestion
// type; an additional "returns" clause overrides it. A lone operand is returned unchanged - no
```

sailingKieler (approved):
Thanks @msujew for this additional feature.

I've just minor remarks, mainly doc and a few nitpicks, see below.
Otherwise it's good to go.

For curiosity: Does the infix operator have measurable performance benefits?

internal/grammar/infix_validator_test.go
Lotes: A question that was more in the background of my mind: would it not be nicer not to block a fixed identifier and generate some not colliding identifier when needed?

ssmifi (approved):
From me as well. Thanks for the answers and the algo. :)

internal/grammar/infix_validator_test.go
msujew: I would argue such a use case is fairly rare + I wouldn't want to change generated output just to avoid collisions. Bad practice IMO.

Lotes: ok, I can live with this :)

msujew:
> For curiosity: Does the infix operator have measurable performance benefits?

Minor only. Before #142, it was more. Now it's a 10% performance improvement for the arithmetics parser benchmark.

PR #135, opened 2026-07-27
Remove TypeId from token

msujew:
The `Token.TypeId` field is an 8-byte field that just allows us to skip a memory lookup. This change removes the field in favor of `Token.Type.Id`, which drops `Token` memory allocation from 64-bytes to 56-bytes (1/8th). Does not reduce parsing performance in any measurable way.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -152,7 +152,9 @@ func (p *ParserState) Consume(tokenType *core.TokenType) *core.Token {
 		return nil
 	}
 	current := p.LA(1)
-	if !tokenType.Matches(current.Type) {
+	// Fast path for the common exact-type match; Matches is a closure call and
+	// only needed for token groups (real tokens never carry a group TypeId).
+	if current.Type.Id != tokenType.Id && !tokenType.Matches(current.Type) {
 		if current.Type == core.EOF {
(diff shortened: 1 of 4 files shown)


parser/parser.go
```diff
@@ -152,7 +152,9 @@ func (p *ParserState) Consume(tokenType *core.TokenType) *core.Token {
 		return nil
 	}
 	current := p.LA(1)
-	if !tokenType.Matches(current.Type) {
+	// Fast path for the common exact-type match; Matches is a closure call and
+	// only needed for token groups (real tokens never carry a group TypeId).
+	if current.Type.Id != tokenType.Id && !tokenType.Matches(current.Type) {
```
Lotes, line 157: Is the first condition not part of `TokenType.Matches(TokenType)`?

OK, this makes cases faster where a list of Ids is compared... but it makes this line a bit more complex... on the other hand, we could have done this before this change, or? `tokenType.TypeId != current.TypeId`

msujew: It's a minor optimization for the default path (speeds up parsing by 1-2%). But yes, we could have done this before as well. Generally `token.TypeId == token.Type.Id`.

Lotes (approved):

PR #136, opened 2026-08-06
Add semantic tokens support

### LINKED ISSUE #95: [LSP] Semantic token provider
Even though it's a semantic service, we should provide similar utilities as [Langium does](https://github.com/eclipse-langium/langium/blob/main/packages/langium/src/lsp/semantic-token-provider.ts) to make it easier for adopters to provide semantic tokens for their languages.

msujew:
Closes https://github.com/TypeFox/fastbelt/issues/95

Provides the infrastructure to implement semantic tokens support for any language. Comes with:
* Testing infrastructure using `AssertSemanticTokens`.
* Options to easily provide custom token types/modifier via `ExtendableSemanticTokensLegendProvider`.
* Composable semantic tokens data via `SemanticTokensBuilder`.
* Simple semantic token highlighting strategies via `TokenHighlightingStrategy` and `TokenBasedSemanticTokensProvider`.
* A semantic token implementation for the grammar language.

### DIFF
--- a/internal/grammar/semantic_tokens.go
+++ b/internal/grammar/semantic_tokens.go
@@ -1,0 +1,55 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+package grammar
+
+import (
+	"context"
+
+	core "typefox.dev/fastbelt"
+	"typefox.dev/fastbelt/server"
+)
+
+var legendProvider = server.NewExtendableSemanticTokensLegendProvider()
+
+type GrammarTokenHighlightingStrategy struct{}
+
+func NewGrammarTokenHighlightingStrategy() server.TokenHighlightingStrategy {
+	return &GrammarTokenHighlightingStrategy{}
+}
+
+func (s *GrammarTokenHighlightingStrategy) Highlight(ctx context.Context, token core.Token, accept server.TokenHighlightingStrategyAcceptor) {
+	switch token.Kind {
+	case Grammar_Name_ID:
+		accept(legendProvider.Namespace(), 0)
+	case Interface_Name_ID,
+		Interface_Extends_ID_0,
+		Interface_Extends_ID_1:
+		accept(legendProvider.Interface(), 0)
+	case ParserRule_Name_ID,
+		Token_Name_ID,
+		CompositeRule_Name_ID,
+		RuleCall_Rule_ID,
+		TokenGroup_Name_ID,
+		TokenGroup_TokenRefs_ID:
+		accept(legendProvider.Function(), 0)
+	case Field_Name_ID,
+		Assignment_Property_ID,
+		Action_Property_ID:
+		accept(legendProvider.Property(), 0)
+	case PrimitiveType_Type_bool,
+		PrimitiveType_Type_composite,
+		PrimitiveType_Type_string,
+		SimpleType_Type_ID,
+		ReferenceType_Type_ID,
+		CrossRef_Type_ID,
+		ParserRule_ReturnType_ID,
+		Action_Type_ID,
+		Action_current:
+		accept(legendProvider.Type(), 0)
+	case Token_Type_comment,
+		Token_Type_hidden:
+		accept(legendProvider.Modifier(), 0)
+	}
+}
--- a/internal/grammar/semantic_tokens.go
+++ b/internal/grammar/semantic_tokens.go
@@ -23,6 +23,8 @@ func (s *GrammarTokenHighlightingStrategy) Highlight(ctx context.Context, token 
 	switch token.Kind {
 	case Grammar_Name_ID:
 		accept(legendProvider.Namespace(), 0)
+	case Keyword_Value_StringLiteral:
+		accept(legendProvider.String(), 0)
 	case Interface_Name_ID,
 		Interface_Extends_ID_0,
 		Interface_Extends_ID_1:
@@ -31,6 +33,7 @@ func (s *GrammarTokenHighlightingStrategy) Highlight(ctx context.Context, token 
 		Token_Name_ID,
 		CompositeRule_Name_ID,
 		RuleCall_Rule_ID,
+		InfixRule_Name_ID,
 		TokenGroup_Name_ID,
 		TokenGroup_TokenRefs_ID:
 		accept(legendProvider.Function(), 0)
--- a/internal/grammar/semantic_tokens.go
+++ b/internal/grammar/semantic_tokens.go
@@ -48,11 +48,14 @@ func (s *GrammarTokenHighlightingStrategy) Highlight(ctx context.Context, token 
 		ReferenceType_Type_ID,
 		CrossRef_Type_ID,
 		ParserRule_ReturnType_ID,
+		InfixRule_ReturnType_ID,
 		Action_Type_ID,
 		Action_current:
 		accept(legendProvider.Type(), 0)
 	case Token_Type_comment,
-		Token_Type_hidden:
+		Token_Type_hidden,
+		PrecedenceGroup_Associativity_left,
+		PrecedenceGroup_Associativity_right:
 		accept(legendProvider.Modifier(), 0)
 	}
 }
--- a/internal/grammar/semantic_tokens_test.go
+++ b/internal/grammar/semantic_tokens_test.go
@@ -35,12 +35,14 @@ token <|function:ID|>: /[a-zA-Z_][a-zA-Z0-9_]*/;
 `
 
 	doc := fixture.ParseURI(grammarText, "file:///semantic.fb")
-	doc.AssertNoParseErrors().
-		AssertSemanticTokens("namespace", legendProvider.Namespace(), 0).
-		AssertSemanticTokens("interface", legendProvider.Interface(), 0).
-		AssertSemanticTokens("function", legendProvider.Function(), 0).
-		AssertSemanticTokens("property", legendProvider.Property(), 0).
-		AssertSemanticTokens("type", legendProvider.Type(), 0).
-		AssertSemanticTokens("modifier", legendProvider.Modifier(), 0).
-		AssertSemanticTokens("comment", legendProvider.Comment(), 0)
+	doc.AssertNoParseErrors()
+	semanticTokens := doc.ExpectSemanticTokens()
+	semanticTokens.
+		Assert("namespace", legendProvider.Namespace(), 0).
+		Assert("interface", legendProvider.Interface(), 0).
+		Assert("function", legendProvider.Function(), 0).
+		Assert("property", legendProvider.Property(), 0).
+		Assert("type", legendProvider.Type(), 0).
+		Assert("modifier", legendProvider.Modifier(), 0).
+		Assert("comment", legendProvider.Comment(), 0)
 }
--- a/internal/grammar/semantic_tokens_test.go
+++ b/internal/grammar/semantic_tokens_test.go
@@ -1,0 +1,46 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+package grammar
+
+import (
+	"testing"
+
+	"typefox.dev/fastbelt/test"
+)
+
+func TestSemanticTokensIntegration(t *testing.T) {
+	fixture := test.New(t, CreateServices())
+
+	grammarText := `<|comment:// Grammar for semantic token testing|>
+grammar <|namespace:Test|>;
+
+interface <|interface:Expression|> {}
+interface <|interface:BinaryExpression|> extends <|interface:Expression|> {
+	<|property:Left|> <|type:Expression|>
+	<|property:Operator|> <|type:string|>
+	<|property:Right|> <|type:Expression|>
+}
+
+<|function:Addition|> returns <|type:Expression|>:
+	<|function:Primary|>
+	({<|type:BinaryExpression|>.<|property:Left|>=<|type:current|>}
+		<|property:Operator|>=("+" | "-") <|property:Right|>=<|function:Primary|>)*
+<|function:Primary|> returns <|type:Expression|>:
+	<|property:Operator|>=<|function:ID|>
+
+token <|function:ID|>: /[a-zA-Z_][a-zA-Z0-9_]*/;
+<|modifier:hidden|> token <|function:WS|>: /[ \n\r\t]+/;
+`
+
+	doc := fixture.ParseURI(grammarText, "file:///semantic.fb")
+	doc.AssertNoParseErrors().
+		AssertSemanticTokens("namespace", legendProvider.Namespace(), 0).
+		AssertSemanticTokens("interface", legendProvider.Interface(), 0).
+		AssertSemanticTokens("function", legendProvider.Function(), 0).
+		AssertSemanticTokens("property", legendProvider.Property(), 0).
+		AssertSemanticTokens("type", legendProvider.Type(), 0).
+		AssertSemanticTokens("modifier", legendProvider.Modifier(), 0).
+		AssertSemanticTokens("comment", legendProvider.Comment(), 0)
+}
--- a/internal/grammar/semantic_tokens_test.go
+++ b/internal/grammar/semantic_tokens_test.go
@@ -29,6 +29,9 @@ interface <|interface:BinaryExpression|> extends <|interface:Expression|> {
 		<|property:Operator|>=("+" | "-") <|property:Right|>=<|function:Primary|>)*
 <|function:Primary|> returns <|type:Expression|>:
 	<|property:Operator|>=<|function:ID|>
+infix <|function:Binary|> on <|function:Primary|> returns <|type:Expression|>:
+	<|modifier:left|> "*" | "/"
+	> <|modifier:right|> "+" | "-";
 
 token <|function:ID|>: /[a-zA-Z_][a-zA-Z0-9_]*/;
 <|modifier:hidden|> token <|function:WS|>: /[ \n\r\t]+/;
--- a/server/semantic_tokens_builder.go
+++ b/server/semantic_tokens_builder.go
@@ -1,0 +1,129 @@
+// Copyright 2026 TypeFox GmbH
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+
+package server
+
+import (
+	"unicode/utf16"
+	"unicode/utf8"
+
+	core "typefox.dev/fastbelt"
+)
+
+type SemanticTokensBuilder interface {
+	Data() []uint32
+	Push(textRange core.TextRange, tokenType, tokenModifiers uint32)
+}
+
+func NewSemanticTokensBuilder(text string, tokenCount int) SemanticTokensBuilder {
+	return &semanticTokensBuilder{
+		// Preallocate the data with the maximum possible length
+		// Each token potentially contributes 5 uint32 values
+		data: make([]uint32, 0, tokenCount*5),
+		text: text,
+	}
+}
+
+type semanticTokensBuilder struct {
+	// data is a slice of uint32 values representing the semantic tokens data in the LSP format.
+	// Each token is represented by five consecutive values:
+	// - deltaLine, token line number, relative to the previous token,
+	// - deltaStart, token start character, relative to the previous token,
+	// - length, the length of the token,
+	// - tokenType, the token type index,
+	// - tokenModifiers, the token modifiers bitset.
+	data        []uint32
+	text        string
+	cursor      int
+	prevLine    int
+	prevChar    int
+	currentLine int
+	currentChar int
+	// lineBreaks is reused across push calls to avoid per-token allocations
+	lineBreaks []int
+}
+
+func (tokenData *semanticTokensBuilder) Data() []uint32 {
+	return tokenData.data
+}
+
+func (tokenData *semanticTokensBuilder) Push(textRange core.TextRange, typeIndex, modifierIndex uint32) {
+	textLen := len(tokenData.text)
+	tokenStart := int(textRange.Start)
+	tokenEnd := int(textRange.End)
+	cursor := tokenData.cursor
+	currentLine := tokenData.currentLine
+	currentChar := tokenData.currentChar
+	startLine, startChar := 0, 0
+	// Count line breaks within the token range as necessary
+	// We need to emit multiple tokens if the token spans multiple lines
+	lineBreaks := tokenData.lineBreaks[:0]
+	// Advance the cursor up to the end of the token
+	for tokenEnd > cursor {
+		if cursor >= textLen {
+			break
+		} else if cursor == tokenStart {
+			startLine = currentLine
+			startChar = currentChar
+		}
+		if c := tokenData.text[cursor]; c < utf8.RuneSelf {
+			// ASCII fast path: one byte, one UTF-16 code unit
+			if c == '\n' {
+				// Record the line break character position
+				if cursor >= tokenStart {
+					lineBreaks = append(lineBreaks, currentChar)
+				}
+				// New line, reset currentChar and increment currentLine
+				currentLine++
+				currentChar = 0
+			} else {
+				currentChar++
+			}
+			cursor++
+			continue
+		}
+		rune, size := utf8.DecodeRuneInString(tokenData.text[cursor:])
+		// Advance column by the number of UTF-16 code units for the rune
+		// (newlines are ASCII, so this rune can never be one)
+		currentChar += utf16.RuneLen(rune)
+		// Advance cursor by the byte size of the rune
+		cursor += size
+	}
+	tokenData.lineBreaks = lineBreaks
+	lineDelta := uint32(startLine - tokenData.prevLine)
+	charDelta := uint32(startChar)
+	if lineDelta == 0 {
+		// If the token is on the same line as the previous token, calculate the character delta
+		charDelta -= uint32(tokenData.prevChar)
+	}
+	if len(lineBreaks) == 0 {
+		// Token is on a single line, emit it directly
+		length := uint32(currentChar - startChar)
+		tokenData.data = append(tokenData.data, lineDelta, charDelta, length, typeIndex, modifierIndex)
+		// Update the previous character position for the next token
+		tokenData.prevChar = startChar
+	} else {
+		// Token spans multiple lines, emit a token for each line segment
+		// First segment: from startChar to the first line break
+		length := uint32(lineBreaks[0] - startChar)
+		tokenData.data = append(tokenData.data, lineDelta, charDelta, length, typeIndex, modifierIndex)
+		// Subsequent segments: from each line break to the next line break
+		for i := 1; i < len(lineBreaks); i++ {
+			// always use the full length of the line
+			length = uint32(lineBreaks[i])
+			// Note: lineDelta is always 1, since each segment is on a new line
+			// charDelta is always 0, since we are starting at the beginning of the line
+			tokenData.data = append(tokenData.data, 1, 0, length, typeIndex, modifierIndex)
+		}
+		// Last segment: from the start of the last line to the end of the token
+		length = uint32(currentChar)
+		tokenData.data = append(tokenData.data, 1, 0, length, typeIndex, modifierIndex)
+		tokenData.prevChar = 0
+	}
+	// Update the data for the next token
+	tokenData.cursor = cursor
+	tokenData.prevLine = currentLine
+	tokenData.currentLine = currentLine
+	tokenData.currentChar = currentChar
+}
--- a/server/semantic_tokens_builder.go
+++ b/server/semantic_tokens_builder.go
@@ -11,15 +11,33 @@ import (
 	core "typefox.dev/fastbelt"
 )
 
+// SemanticTokensBuilder defines the interface for building semantic tokens data in the LSP format.
+// It provides methods to push individual tokens and retrieve the final data slice.
+//
+// It is recommended to build the semantic tokens data by using the [TokenBasedSemanticTokensProvider] and its
+// associated [TokenHighlightingStrategy] implementations.
 type SemanticTokensBuilder interface {
+	// Data returns the final semantic tokens data slice in the LSP format.
+	// The LSP data slice is a flat array of uint32 values, where each token is represented by five consecutive values:
+	// (1) deltaLine: token line number, relative to the previous token,
+	// (2) deltaStart: token start character, relative to the previous token,
+	// (3) length: the length of the token,
+	// (4) tokenType: the token type index in the legend, and
+	// (5) tokenModifiers: the token modifiers bitset in the legend.
 	Data() []uint32
+	// Push adds a new token to the semantic tokens data.
+	// Tokens should be pushed in the order they appear in the document, as the LSP
+	// data is based on the offset of the previous token.
+	// Due to this, this method is not thread-safe.
 	Push(textRange core.TextRange, tokenType, tokenModifiers uint32)
 }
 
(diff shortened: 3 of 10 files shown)


test/doc_fixture_lsp.go
```diff
@@ -420,3 +420,74 @@ func findSymbolAtRange(symbols []lsp.DocumentSymbol, targetRange lsp.Range) *lsp
...
+// Returns the [Doc] for chaining.
+func (d *Doc) AssertSemanticTokens(label string, expectedType uint32, expectedModifiers uint32) *Doc {
+	d.fixture.t.Helper()
+
+	ranges := d.markerRanges(label)
+	if len(ranges) == 0 {
+		d.fixture.t.Fatalf("fbtest: no marker with label %q", label)
+	}
+
+	semanticTokensProvider := service.MustGet[server.SemanticTokensProvider](d.fixture.sc)
```
Lotes, line 428: Would a cache make sense? AFAICS this is recomputed for each assertion.

test/doc_fixture_lsp.go
```diff
@@ -420,3 +420,74 @@ func findSymbolAtRange(symbols []lsp.DocumentSymbol, targetRange lsp.Range) *lsp
...
+			// Same line, column is relative to previous token
+			column += result.Data[i+1]
+		} else {
+			// New line, column is absolute
+			column = result.Data[i+1]
+		}
+		tokens = append(tokens, semanticToken{line, column, result.Data[i+2], result.Data[i+3], result.Data[i+4]})
+	}
+
+	for _, rng := range ranges {
```
Lotes, line 479: Binary Search? On the other hand it is just for tests…

Lotes (commented):
I have found 2 things.

test/doc_fixture_lsp.go
msujew: I thought about that as well, but I can't imagine this ever becoming a performance bottleneck. I would keep it as is for now. An alternative API that I've considered (similar to the diagnostics assertion API):

```go
semanticTokens := doc.ExpectSemanticTokens()
semanticTokens.Assert("function", x, y)
semanticTokens.Assert("variable", u, v)
```

WDYT? We can then cache the value on `semanticTokens`.

test/doc_fixture_lsp.go
msujew: Yeah, would probably be over-engineered.

test/doc_fixture_lsp.go
Lotes: Feels better :-)

msujew: Implemented in the latest commit.

internal/grammar/semantic_tokens_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+token <|function:ID|>: /[a-zA-Z_][a-zA-Z0-9_]*/;
+<|modifier:hidden|> token <|function:WS|>: /[ \n\r\t]+/;
+`
+
+	doc := fixture.ParseURI(grammarText, "file:///semantic.fb")
+	doc.AssertNoParseErrors()
+	semanticTokens := doc.ExpectSemanticTokens()
+	semanticTokens.
+		Assert("namespace", legendProvider.Namespace(), 0).
```
Lotes, line 41: I stumbled over this already the last time, but did not mention it:
`legendProvider` seems a singleton in the global scope?
I had problems to find it. What if we want to override the legend provider?

Lotes: Does this even make sense xD?
Meanwhile I found the singleton.

Lotes (approved):
Thanks! Looks good. Just some little questions about the legend provider.

internal/grammar/semantic_tokens_test.go
msujew: Each language has its own `legendProvider`. Note that it's declared in `internal/grammar`, which is inaccessible to adopters. No need to override it. Adopters should declare their own and bind it. That's how the interface is designed.

internal/grammar/semantic_tokens_test.go
```diff
@@ -35,12 +35,14 @@ token <|function:ID|>: /[a-zA-Z_][a-zA-Z0-9_]*/;
...
+		Assert("interface", legendProvider.Interface(), 0).
+		Assert("function", legendProvider.Function(), 0).
+		Assert("property", legendProvider.Property(), 0).
+		Assert("type", legendProvider.Type(), 0).
+		Assert("modifier", legendProvider.Modifier(), 0).
+		Assert("comment", legendProvider.Comment(), 0)
```
sailingKieler, line 50: Are these trailing `.` common practice? Never seen it like this

server/semantic_tokens_legend.go
```diff
@@ -0,0 +1,237 @@
...
+//
+//	// Declaring a new legend provider with custom token types and modifiers
+//	var legendProvider = server.NewExtendableSemanticTokensLegendProvider()
+//	// Returns the index of the new token type within the legend
+//	var extraType = legendProvider.AddType("extraTokenType")
+//	// Returns the bit value of the new token modifier within the legend
+//	var extraModifier = legendProvider.AddModifier("extraTokenModifier")
+//	// Register within the service container
+//	service.Put[server.SemanticTokensLegendProvider](sc, legendProvider)
+type ExtendableSemanticTokensLegendProvider interface {
```
sailingKieler, line 34: It took my a while until I got what the contribution of this is.

I suggest to name `DefaultSemanticTokensLegendProvider`, not because that's our typical scheme but because it's configured with the default token types being used in VS Code's standard UI themes.

Also mention in the docs that all its additional methods are just sugar for easy access of the default token type ids.

server/semantic_tokens_builder.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (tokenData *semanticTokensBuilder) Push(textRange core.TextRange, typeIndex, modifierIndex uint32) {
+	textLen := len(tokenData.text)
+	tokenStart := int(textRange.Start)
+	tokenEnd := int(textRange.End)
+	cursor := tokenData.cursor
+	currentLine := tokenData.currentLine
+	currentChar := tokenData.currentChar
+	startLine, startChar := 0, 0
```
sailingKieler: **Data corruption via `uint32` underflow**: `Push` only assigns `startLine`/`startChar` from their `0, 0` default inside the `cursor == tokenStart` branch (lines 66-69). If a call's range starts at/behind the already-consumed cursor — an overlapping or out-of-order range, or a range past `textLen` — that branch never runs, and `startLine`/`startChar` stay `0`, corrupting the emitted delta via unsigned underflow.

Reproduced directly against this code:
```go
text := "0123456789abcdef"
b := NewSemanticTokensBuilder(text, 4)
b.Push(core.NewTextRange(5, 10), 1, 0) // pushes "56789"
b.Push(core.NewTextRange(2, 8), 1, 0)  // overlaps the previous range
fmt.Println(b.Data())
// [0 5 5 1 0 0 4294967291 10 1 0]
//        ^ deltaStart wrapped to uint32(-5)
```

Today's only caller (`TokenBasedSemanticTokensProvider`) always passes sorted, non-overlapping, in-bounds ranges (tokens from the lexer, comments merged via a strict `<` tie-break), so this isn't reachable through this PR's own code paths. But `SemanticTokensBuilder`/`NewSemanticTokensBuilder` are exported with no doc stating the "strictly increasing, non-overlapping, in-bounds" precondition — a future strategy or external caller that violates it gets silently corrupted `SemanticTokens.Data` sent to the client, not an error. Worth at least documenting the precondition, or asserting it and returning/panicking with a clear message instead of silently underflowing.

server/semantic_tokens_provider.go
```diff
@@ -0,0 +1,109 @@
...
+	strategy TokenHighlightingStrategy
+}
+
+// NewTokenBasedSemanticTokensProvider creates a new instance of [TokenBasedSemanticTokensProvider] with the given [TokenHighlightingStrategy].
+func NewTokenBasedSemanticTokensProvider(sc *service.Container, strategy TokenHighlightingStrategy) SemanticTokensProvider {
+	return &TokenBasedSemanticTokensProvider{sc: sc, strategy: strategy}
+}
+
+func (p *TokenBasedSemanticTokensProvider) HandleSemanticTokensFullRequest(ctx context.Context, params *lsp.SemanticTokensParams) (*lsp.SemanticTokens, error) {
+	documentManager := service.MustGet[workspace.DocumentManager](p.sc)
```
sailingKieler, line 73: Brought up by Claude, basically a duplicate of a remark in `server.go`.

**Capability/handler pairing is unenforced, and the panic path is unrecovered.** `Initialize` (server.go:57) advertises the `SemanticTokensProvider` server capability based on whether a `SemanticTokensLegendProvider` is registered, but this handler requires a *different* service, `SemanticTokensProvider` (server.go:490). The doc comments on both interfaces say "must be registered together," but nothing enforces it — `internal/grammar/services.go` happens to register both via two independent `service.Put` calls.

If a future language package registers only one of the two, the mismatch surfaces as a panic here: `service.MustGet[SemanticTokensLegendProvider]` panics if that service is missing. Traced the call chain and confirmed there's no `recover()` anywhere between this handler and the connection's dispatcher goroutine (`workspace/lock.go`'s `DefaultLock.Read`, `typefox.dev/lsp`'s `ServerHandler`, and `golang.org/x/exp/jsonrpc2`'s `deliverMessages` all lack one) — so this isn't a "that request fails" bug, it's a "the whole language server process crashes" bug, for every open document/connection.

Given the blast radius, I'd suggest making the pairing a hard invariant checked at startup (e.g. in `Container.Seal()` or a dedicated validation step) rather than relying on two docstrings and hoping every registrant reads them.

server/server.go
```diff
@@ -471,7 +483,22 @@ func (s *DefaultLanguageServer) SelectionRange(ctx context.Context, params *lsp.
 	return nil, nil
 }
 func (s *DefaultLanguageServer) SemanticTokensFull(ctx context.Context, params *lsp.SemanticTokensParams) (*lsp.SemanticTokens, error) {
-	return nil, nil
+	lock, err := service.Get[workspace.Lock](s.sc)
+	if err != nil {
+		return nil, err
+	}
+	tokensProvider, err := service.Get[SemanticTokensProvider](s.sc)
```
sailingKieler, line 493: Corresponds to the remark **Capability/handler pairing is unenforced, and...** in `semantic_tokens_provider.go` above:

**Behavior regression for every language server that doesn't register a `SemanticTokensProvider`.** `service.Get[SemanticTokensProvider](s.sc)` returns a hard error when that service is absent, as its presence is never evaluated.

Checked the rest of the repo — only `internal/grammar` registers `SemanticTokensProvider`/`SemanticTokensLegendProvider`. None of `examples/statemachine`, `examples/arithmetics`, `internal/languages/lookahead`, `internal/languages/completion`, or `internal/languages/token_groups` do. For all of those, a client that (incorrectly, or via a race with `Initialize`) sends this request now gets a visible JSON-RPC error where it previously got silent `null`. Also, the error returned is a bare `fmt.Errorf("service ... not found", ...)`, not an LSP-shaped error (e.g. `MethodNotFound`), so even the "client shouldn't have called this" case surfaces as an opaque internal error.

No existing test exercises `DefaultLanguageServer.SemanticTokensFull` itself (the new `ExpectSemanticTokens` test helper calls `SemanticTokensProvider.HandleSemanticTokensFullRequest` directly, bypassing this method entirely), so this regression wouldn't be caught by CI.

server/semantic_tokens_provider.go
```diff
@@ -0,0 +1,109 @@
...
+			if !added {
+				tokenBuilder.Push(token.Range, tokenType, tokenModifier)
+				added = true
+			} else {
+				errorRanges = append(errorRanges, token.Range)
+			}
+		})
+	}
+	// Report any tokens that were highlighted multiple times for the same range
+	if len(errorRanges) > 0 {
```
sailingKieler: I agree with Claude's finding of the undocumented runtime constraint:

**A single double-classified token discards the entire document's semantic tokens.** If `strategy.Highlight` ever calls `accept` twice for the same token, the range is recorded in `errorRanges`, but the function still returns `nil, error` for the *whole request* — throwing away every correctly-computed token in the file, not just mis-highlighting the one offending token.

Neither `TokenHighlightingStrategy.Highlight`'s doc comment nor `TokenHighlightingStrategyAcceptor`'s states a "call `accept` at most once" contract, so this is discoverable only by reading this file's source. A single overlapping `case` in a future strategy's `switch` (easy to introduce, e.g. two case values that both legitimately match) turns into "no highlighting at all for this file, on every request" rather than a scoped, debuggable failure.

server/semantic_tokens_provider.go
```diff
@@ -0,0 +1,109 @@
...
+	commentTypeIndex := slices.Index(tokenTypes, string(lsp.CommentType))
+	tokenBuilder := NewSemanticTokensBuilder(doc.TextDoc.Text(nil), totalLen)
+	var errorRanges []core.TextRange
+	commentIndex := 0
+	for _, token := range tokens {
+		for commentTypeIndex != -1 &&
+			commentIndex < len(comments) &&
+			comments[commentIndex].Range.Start < token.Range.Start {
+			// Add all comments that precede the current token
+			tokenBuilder.Push(comments[commentIndex].Range, uint32(commentTypeIndex), 0)
```
sailingKieler: Also a good finding by Claude, see below.
What do think about an additional optional `commentStrategy` in `TokenBasedSemanticTokensProvider` that can be contributed during initialization?
If given that strategy is applied, otherwise the current behavior is executed.
That would enable fine-grained formatting of comments like java/js/go doc, I'm in favor of adding the (minimal) foundations here already.

**Comment highlighting bypasses `TokenHighlightingStrategy` entirely.** Every comment is pushed with the same hardcoded `(commentTypeIndex, 0)` pair, found via a linear scan for `lsp.CommentType` in the legend — the strategy never sees comment tokens at all. That forecloses functionality the token model already supports: `TokenType.Group` lets a grammar register multiple distinct comment-like token kinds (line vs. block vs. doc comment) that all currently get flattened into "comment, no modifiers" here.

Compare with the established pattern elsewhere in this package — `DocumentSymbolFilter`/`FoldingRangeFilter` give the caller small, targeted hooks for exactly this kind of per-language customization (e.g. `FoldingRangeFilter.IncludeLastFoldingLineForComment()`). `TokenBasedSemanticTokensProvider` gives zero hooks for comments, which is a step back from that precedent. Not blocking, but worth a follow-up if doc-comments (with the `documentation` modifier) or distinct comment types are ever wanted.

server/semantic_tokens_legend.go
```diff
@@ -0,0 +1,237 @@
...
+var _readonly uint32 = 1 << add("readonly", &defaultTokenModifiers)
+var _static uint32 = 1 << add("static", &defaultTokenModifiers)
+var _deprecated uint32 = 1 << add("deprecated", &defaultTokenModifiers)
+var _abstract uint32 = 1 << add("abstract", &defaultTokenModifiers)
+var _async uint32 = 1 << add("async", &defaultTokenModifiers)
+var _modification uint32 = 1 << add("modification", &defaultTokenModifiers)
+var _documentation uint32 = 1 << add("documentation", &defaultTokenModifiers)
+var _defaultLibrary uint32 = 1 << add("defaultLibrary", &defaultTokenModifiers)
+
+type extendableSemanticTokensLegendProvider struct {
```
sailingKieler, line 166: Maybe an edge case. However, I think its worth explaining where/at which point at runtime to register such extra token types & modifiers.
May it be worth to panic if more items are registered after the container is sealed?

**`AddType`/`AddModifier`/`Legend()` aren't safe for concurrent use, despite the type's own doc comment recommending runtime extension.** `add()` (line 119-123) is a non-atomic read-then-append (`len` then `append`) on `d.types`/`d.modifiers` with no mutex. Two concurrent `AddType` calls on the same provider instance can read the same length before either appends, silently handing out the *same* index to two different token-type names. Separately, `HandleSemanticTokensFullRequest` calls `Legend()` (which copies `d.types`/`d.modifiers`) under `Lock.Read`, and `workspace/lock.go`'s reader count confirms multiple readers run concurrently — so an `AddType` call racing with an in-flight `Legend()` call is an unsynchronized slice read/write, a `go test -race` hit (possible `slice bounds out of range` panic or a truncated legend sent to the client).

This is latent today (no in-repo caller does concurrent registration), but the doc comment on `ExtendableSemanticTokensLegendProvider` explicitly frames `AddType`/`AddModifier` as safe to call "at runtime," which reads as an invitation for exactly this hazard once a plugin/lazy-registration use case shows up.

server/semantic_tokens_legend.go
```diff
@@ -0,0 +1,237 @@
...
+	Event() uint32
+	// Function returns the index of the "function" token type within the legend.
+	Function() uint32
+	// Method returns the index of the "method" token type within the legend.
+	Method() uint32
+	// Macro returns the index of the "macro" token type within the legend.
+	Macro() uint32
+	// Keyword returns the index of the "keyword" token type within the legend.
+	Keyword() uint32
+	// Modifier returns the index of the "modifier" token type within the legend.
```
sailingKieler, line 67: Also a very good point.
Please, at least, extend the doc of this function to make clear that it refers to something very different than `AddModifier`.

**Naming footgun**: `Modifier()` returns the index of the "modifier" token *type* (same shape as `Type()`, `Keyword()`, `Function()`, etc.), while `Declaration()`, `Readonly()`, `Deprecated()`, and the rest of the block below return pre-shifted modifier *bit* values. Both groups sit in the same flat interface and return a bare `uint32`, distinguished only by doc-comment prose — `Modifier()` is also the most natural-sounding name for "the modifier bit value" to a strategy author skimming the interface, even though that's not what it does.

Not a bug in this PR (`internal/grammar/semantic_tokens.go` uses `Modifier()` correctly, as a token type for the `hidden`/`comment` grammar keywords), but worth considering a rename (e.g. `ModifierType()`) or splitting the two groups into separate interfaces/methods to make the type-vs-bitset distinction visible to the type system rather than relying on careful reading.

test/doc_fixture_lsp.go
```diff
@@ -420,3 +420,88 @@ func findSymbolAtRange(symbols []lsp.DocumentSymbol, targetRange lsp.Range) *lsp
...
+			URI: lsp.DocumentURI(d.Document.URI.DocumentURI()),
+		},
+	}
+	result, err := semanticTokensProvider.HandleSemanticTokensFullRequest(d.fixture.ctx, params)
+	if err != nil {
+		d.fixture.t.Fatalf("fbtest: HandleSemanticTokensFullRequest returned error: %v", err)
+	}
+	var tokens []semanticToken
+	var line, column uint32
+	for i := 0; i+4 < len(result.Data); i += 5 {
```
sailingKieler, line 442: Good catch, although we probably won't call this test method with an empty document input.
However, I might bite us in some more complex application project test scenario.

**Nil dereference when the document has no tokens and no comments.** `HandleSemanticTokensFullRequest` explicitly returns `nil, nil` when `totalLen := len(tokens)+len(comments)` is `0` (and also when the document isn't found) — see `semantic_tokens_provider.go:60-61`. This helper only checks `err != nil` before immediately doing `len(result.Data)` a few lines below, so any test calling `ExpectSemanticTokens()` against an empty file, a whitespace-only file, or an unknown URI panics with a nil-pointer dereference instead of failing cleanly via `t.Fatalf`.

server/server.go
```diff
@@ -44,19 +44,30 @@ func (s *DefaultLanguageServer) Initialize(ctx context.Context, params *lsp.Para
 		return nil, err
 	}
 	workspaceFolders.Value = params.WorkspaceFolders
-	var triggerChars []string
-	if triggers, err := service.Get[CompletionTriggers](s.sc); err == nil && triggers != nil {
-		triggerChars = triggers.TriggerCharacters()
+	var completionOptions *lsp.CompletionOptions
+	if completionProvider, err := service.Get[CompletionProvider](s.sc); err == nil && completionProvider != nil {
```
sailingKieler: Not sure whether I got every detail in this remark, at least the consistency is worth to look at another time I guess.

**Typed-nil hazard, and inconsistent with this method's own convention.** Every other optional capability below this block uses `service.Has[XProvider](s.sc)` — a pure presence check (see `DefinitionProvider`, `DocumentSymbolProvider`, `FoldingRangeProvider`, etc. a few lines down). This new block (and the mirrored `SemanticTokensLegendProvider` block at line 57) instead does `service.Get[T](s.sc)` plus `x != nil`. Since `service.Put` stores whatever value is passed verbatim with no nil-check, a registrant that does `service.Put[CompletionProvider](sc, (*someImpl)(nil))` produces a non-nil interface wrapping a nil pointer — the classic Go typed-nil gotcha — and `err == nil && completionProvider != nil` would be true, advertising a capability that panics on first use. Not currently triggered by any registrant in this repo, but there's no reason for this block to deviate from the `service.Has` pattern the rest of the method already establishes.

server/semantic_tokens_builder.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+type SemanticTokensBuilder interface {
+	Data() []uint32
+	Push(textRange core.TextRange, tokenType, tokenModifiers uint32)
+}
+
+func NewSemanticTokensBuilder(text string, tokenCount int) SemanticTokensBuilder {
+	return &semanticTokensBuilder{
+		// Preallocate the data with the maximum possible length
```
sailingKieler: Found by claude, see below.
However, ignored SL comment tokens aren't seen by the `SemanticTokensBuilder`, are they?

**Preallocation comment doesn't hold for multi-line tokens.** The comment claims this is "the maximum possible length" at `tokenCount*5`, reasoning "each token potentially contributes 5 uint32 values" — but a single token/comment spanning N lines emits N separate 5-tuples (see the multi-line branch in `Push`, lines 106-123). A document with a handful of tokens and one 200-line block comment needs far more than `tokenCount*5` capacity, so `append` ends up growing/reallocating repeatedly on exactly the documents (large files, big block comments) where the preallocation is meant to help. `doc.TextDoc.LineCount()` is O(1) (cached), so factoring line count into the capacity estimate (e.g. `max(tokenCount, lineCount)*5`) is cheap and available at the call site.

internal/grammar/semantic_tokens.go
```diff
@@ -0,0 +1,55 @@
...
+		accept(legendProvider.Property(), 0)
+	case PrimitiveType_Type_bool,
+		PrimitiveType_Type_composite,
+		PrimitiveType_Type_string,
+		SimpleType_Type_ID,
+		ReferenceType_Type_ID,
+		CrossRef_Type_ID,
+		ParserRule_ReturnType_ID,
+		Action_Type_ID,
+		Action_current:
```
sailingKieler: **Possible misclassification**: `Action_current` is the literal `current` keyword (consumed via `p.state.Consume(Keyword_current)` in the generated parser — analogous to `this`/`self`), not a type reference. It's grouped into this case alongside genuine type/cross-reference identifiers (`SimpleType_Type_ID`, `ReferenceType_Type_ID`, `CrossRef_Type_ID`, `ParserRule_ReturnType_ID`, `Action_Type_ID`), so it gets highlighted with the `type` token type. Other structurally-identical keyword literals in the grammar (e.g. `returns`, `extends`) aren't highlighted at all. Was grouping `current` under `Type()` intentional, or did it end up here via copy/paste from the adjacent `Action_Type_ID` line? If unintentional, an editor theme that colors `type` tokens like a class/interface name will render the `current` keyword the same way, which is likely not the intent.

server/semantic_tokens_legend.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	return add(name, &d.types) + uint32(len(defaultTokenTypes))
+}
+
+func (d *extendableSemanticTokensLegendProvider) AddModifier(name string) uint32 {
+	if slices.Contains(defaultTokenModifiers, name) {
+		panic("Cannot add a token modifier that already exists in the default legend: " + name)
+	} else if slices.Contains(d.modifiers, name) {
+		panic("Cannot add a token modifier that already exists in the legend: " + name)
+	} else if (len(d.modifiers) + len(defaultTokenModifiers)) >= 32 {
```
sailingKieler, line 228: Interesting remark by Claude, see subsequently.
Is the max 32 constrained somewhere explained in the docs? Can't find any hint.

**`AddModifier`'s capacity panic is a global aggregate invariant, unlike this file's other panics.** `AddType`'s duplicate-name panics (lines 207-211) are triggered by a single, local, staticaly-inspectable call — easy to attribute and fix. This one instead depends on the *cumulative* total of however many independent `AddModifier` calls have run against the same legend instance across the whole program's history: a perfectly correct individual call can be "the one that broke the budget" pushed over by unrelated, earlier additions, and the panic message doesn't say which caller is responsible or how many slots remain. For a downstream language package accumulating modifiers over time/features, the 32nd add just panics the whole binary with no actionable diagnostic. Worth considering a non-panicking path for this specific case (e.g. return `(uint32, error)`, or expose a `RemainingModifierSlots()` guard callers can check first) while keeping panic for the genuinely local typo case.

server/semantic_tokens_provider.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// NewTokenBasedSemanticTokensProvider creates a new instance of [TokenBasedSemanticTokensProvider] with the given [TokenHighlightingStrategy].
+func NewTokenBasedSemanticTokensProvider(sc *service.Container, strategy TokenHighlightingStrategy) SemanticTokensProvider {
+	return &TokenBasedSemanticTokensProvider{sc: sc, strategy: strategy}
+}
+
+func (p *TokenBasedSemanticTokensProvider) HandleSemanticTokensFullRequest(ctx context.Context, params *lsp.SemanticTokensParams) (*lsp.SemanticTokens, error) {
+	documentManager := service.MustGet[workspace.DocumentManager](p.sc)
+	tokenTypes := service.MustGet[SemanticTokensLegendProvider](p.sc).Legend().TokenTypes
```
sailingKieler: Agree with Claude in this:

**`Legend()` is rebuilt from scratch on every request, and half the result is discarded immediately.** `Legend()` does two `make`+`copy` pairs; this call only uses `.TokenTypes` (line 51), so the `TokenModifiers` slice it also builds is 100% wasted allocation on every single `textDocument/semanticTokens/full` request. The immediately-following `commentTypeIndex := slices.Index(tokenTypes, ...)` (line 63) then does an O(len(tokenTypes)) string-compare scan for a value that's constant after startup (the legend "virtually never changes after startup," per `ExtendableSemanticTokensLegendProvider`'s own doc comment).

On an editing session firing this per keystroke-pause debounce and per scroll of a large file, that's continuous, entirely avoidable allocation + scanning for data that's byte-identical to what `Initialize` already computed once. Worth memoizing the built legend (e.g. `sync.Once`, since extensions can still register via `AddType`/`AddModifier` before first use) and caching `commentTypeIndex` alongside it.

server/semantic_tokens_builder.go
```diff
@@ -0,0 +1,129 @@
...
+package server
+
+import (
+	"unicode/utf16"
+	"unicode/utf8"
+
+	core "typefox.dev/fastbelt"
+)
+
+type SemanticTokensBuilder interface {
```
sailingKieler, line 19: Please add helpful docs to this interface and its methods, in particular explain the semantics of the data returned by `Data()`.

server/semantic_tokens_legend.go
```diff
@@ -0,0 +1,237 @@
...
+	Operator() uint32
+	// Decorator returns the index of the "decorator" token type within the legend.
+	Decorator() uint32
+	// Label returns the index of the "label" token type within the legend.
+	Label() uint32
+	// Namespace returns the index of the "namespace" token type within the legend.
+	Namespace() uint32
+
+	// Declaration returns the bit value of the "declaration" token modifier within the legend.
+	Declaration() uint32
```
sailingKieler: In the face of the remark above:
What about prefixing the names of this and the following funcs with `Mod_`, or `Bit_`?

sailingKieler (commented):
Thanks a lot @msujew for this contribution, too, now I made it through it.

The review done by Claude brought interesting issues up, and I have a few requests for more documentation, see below.

internal/grammar/semantic_tokens_test.go
msujew: If no trailing `.` is provided, the Go lexer will automatically insert `;`, which breaks the fluent API.

server/semantic_tokens_builder.go
msujew: I'll add a comment. The API is not designed to push unsorted tokens (that would require O(n * log(n)) sorting).

server/semantic_tokens_provider.go
msujew: I think it's fine to crash the language server if the registration isn't done correctly. But I'll add a more descriptive message for this.

server/semantic_tokens_legend.go
msujew: > For a downstream language package accumulating modifiers over time/features, the 32nd add just panics the whole binary with no actionable diagnostic.

That use case doesn't make any sense. You can't dynamically register new modifiers to a language client.

I'll add a comment explaining that you can't add more than 32 modifiers.

server/semantic_tokens_legend.go
msujew: > it's configured with the default token types being used in VS Code's standard UI themes.

That's not quite correct. Those are the default token types documented in the LSP.

> I suggest to name DefaultSemanticTokensLegendProvider

The issue with `Default` is that it indicates that it's the "default" semantic token legend provider, which isn't correct. I'm not a big fan of having two different kind of "default" connotations in our code base.

> Also mention in the docs that all its additional methods are just sugar for easy access of the default token type ids.

This is already stated by the documentation.

server/server.go
msujew: I'm confused by this remark. It's correct that clients that call this endpoint without it being advertised in the server capabilities get an error instead of `null`.

server/server.go
msujew: If someone wants to shoot themselves in the foot by doing `service.Put[CompletionProvider](sc, (*someImpl)(nil))`, they are free to do so. Note that this issue would also trigger in the `service.Has` case, so the recommendation doesn't really help at all...

server/semantic_tokens_builder.go
msujew: I would keep it as is it. If we need to resize - so it be.

server/semantic_tokens_provider.go
msujew: I've added additional documentation for this. Note that the user will get a scoped, debuggable failure info, since we return this as part of the error response.

server/semantic_tokens_legend.go
msujew: > May it be worth to panic if more items are registered after the container is sealed?

We can't tell.

Anyway, I've added primitives to make sure it's safe for concurrent access (whoever would use that?).

server/semantic_tokens_legend.go
msujew: I've renamed the other bitset values to use the `Mod` prefix.

server/server.go
sailingKieler: If just a `SemanticTokensLegendProvider` is registered, the LS reports that support of semantic tokens is available during the initialization, correct?
If no `SemanticTokensProvider` is registered, the server will crash here.

What about having the legend provider a member of the `SemanticTokensProvider` similar to the strategy, like this:
```
        service.Put(sc, server.NewTokenBasedSemanticTokensProvider(sc,
            legendProvider,
            NewGrammarTokenHighlightingStrategy(),
        ))
```

That would sort out this potential pitfall.

msujew: Sure, I'm fine with this. My most recent commit moves this into the token provider 👍

PR #142, opened 2026-08-10
Reduce parser memory allocations

msujew:
Implements three parser optimizations that massively reduce the amount of memory allocations required for parsing:
1. Instead of constructing the `FollowSet`, we simply iterate over the parser stack via `FollowSetContains`. The additional loop is much cheaper than the allocations of the new `BitSet`.
2. Within each `AstNodeBase`, allocate 4 tokens worth of token buffer together with the AST node. This ensures that a small token buffer is always available. Allocation only needs to happen when there are more tokens, which is rare for most AST nodes.
3. Instead of allocating AST nodes always when the parser rule is called, this change adjusts the parser generator to only instantiate the AST node, if actually required. This is the largest change, but massively reduces the amount of allocations required for any kind of expression language.

In total, these three optimizations quadruple the parsing performance for the arithmetics language. With it, we go from ~7.3MB/s of parsing speed to ~32MB/s on my machine. Also has benefits on the statemachine language benchmark.

### DIFF
--- a/ast.go
+++ b/ast.go
@@ -82,7 +82,12 @@ type AstNodeBase struct {
 	containerField unique.Handle[string]
 	containerIndex int
 	tokens         []*Token
-	rng            TextRange
+	// tokenBuf is a small preallocated buffer to avoid heap allocations for nodes with few tokens.
+	// It is used as the backing array for the tokens slice when the node has 4 or fewer tokens.
+	// Massively reduces the amount of allocations required for most languages, which increases
+	// the parsing speed by roughly 25% in benchmarks.
+	tokenBuf [4]*Token
+	rng      TextRange
 }
 
 // Document returns the owning document of the node.
@@ -182,13 +187,9 @@ func (node *AstNodeBase) TextRange() TextRange {
 func (node *AstNodeBase) AppendToken(token *Token) {
 	if node != nil && token != nil {
 		if node.tokens == nil {
-			// Most nodes end up with a handful of tokens; start with capacity 4
-			// to avoid the cap 1->2->4 growth reallocations of a bare append.
-			node.tokens = make([]*Token, 1, 4)
-			node.tokens[0] = token
-		} else {
-			node.tokens = append(node.tokens, token)
+			node.tokens = node.tokenBuf[:0]
 		}
+		node.tokens = append(node.tokens, token)
 	}
 }
 
--- a/internal/generator/parser_generator.go
+++ b/internal/generator/parser_generator.go
@@ -107,6 +107,132 @@ type ParserGeneratorContext struct {
 	// counter for unique Go loop labels (loop0, loop1, ...)
 	// Only used when the generator can combine an Alternatives node with its loop
 	loopLabelSeq int
+	// lazyCurrent is active while generating a rule that allocates its `current`
+	// node on demand: at least one path replaces `current` before touching it,
+	// so the eager placeholder allocation would be pure garbage on that path.
+	// Rules without such a path keep the eager allocation and identical output.
+	lazyCurrent bool
+	// currentNil is true while `current` is statically known to be nil on the
+	// emission path of a lazyCurrent rule.
+	currentNil bool
+	// currentType is the rule return type used to materialize a lazy `current`.
+	currentType string
+}
+
+// ensureCurrent materializes a deferred `current` node right before emitted
+// code that reads or mutates it. No-op unless `current` is known to be nil.
+func (c *ParserGeneratorContext) ensureCurrent(node codegen.Node) {
+	if !c.lazyCurrent || !c.currentNil || c.completion {
+		return
+	}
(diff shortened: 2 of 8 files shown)


ast.go
```diff
@@ -82,7 +82,12 @@ type AstNodeBase struct {
 	containerField unique.Handle[string]
 	containerIndex int
 	tokens         []*Token
-	rng            TextRange
+	// tokenBuf is a small preallocated buffer to avoid heap allocations for nodes with few tokens.
+	// It is used as the backing array for the tokens slice when the node has 4 or fewer tokens.
+	// Massively reduces the amount of allocations required for most languages, which increases
+	// the parsing speed by roughly 25% in benchmarks.
+	tokenBuf [4]*Token
```
Lotes, line 89: Question: When APPEND has more than 4 tokens we still have 4 token pointers allocated at every ASTNode, or am I wrong?
(I would prefer the previous implementation in that case.)

Lotes: I mean for cases >4 tokens, is this field not garbage then, that is never collected?

internal/generator/parser_generator.go
```diff
@@ -107,6 +107,132 @@ type ParserGeneratorContext struct {
...
+// materializeCurrentGuard emits a nil-guarded materialization for join points
+// where only some incoming paths allocated `current` (e.g. after a lazy
+// alternatives switch, whose no-viable-alternative arm allocates nothing).
+func (c *ParserGeneratorContext) materializeCurrentGuard(node codegen.Node) {
+	if !c.lazyCurrent || !c.currentNil || c.completion {
+		return
+	}
+	node.AppendLine("if current == nil {")
+	node.Indent(func(n codegen.Node) {
+		n.AppendLine("current = New", c.currentType, "()")
```
Lotes, line 140: `materializeCurrentGuard` and `ensureCurrent` are very similar to each other.
Maybe add a helper function to carry out the common parts.

internal/generator/parser_generator.go
```diff
@@ -107,6 +107,132 @@ type ParserGeneratorContext struct {
 	// counter for unique Go loop labels (loop0, loop1, ...)
 	// Only used when the generator can combine an Alternatives node with its loop
 	loopLabelSeq int
+	// lazyCurrent is active while generating a rule that allocates its `current`
+	// node on demand: at least one path replaces `current` before touching it,
+	// so the eager placeholder allocation would be pure garbage on that path.
+	// Rules without such a path keep the eager allocation and identical output.
+	lazyCurrent bool
```
Lotes, line 114: What is "current" about? It is the "current" inside actions? (`{Group.Elements+=current}`)

ast.go
msujew: How this works:
1. The first time we call `append`, this array is used as the backing array, and one token is appended.
2. Until the first 4 tokens are set `append` always returns the same array pointer.
3. Once a fifth token is added, `tokens` is placed into a new, heap-allocated array, the old array stays in place, but is unused, since adopters cannot interact with it.

Yes, as a consequence it always stays there. But most nodes don't have more than 4 tokens. I did a quick test and it turns out that in >99% of AST nodes (for statemachine & arithmetics), the runtime doesn't need to allocate the new array. My data:

**Arithmetics** (benchmark document, 1,119 nodes - 99.9% inline, 0.1% spilled):

| Tokens | Nodes | |
|---|---|---|
| 1 | 915 | number literals, binary expressions (the operator) |
| 2 | 1 | |
| 3 | 101 | |
| 4 | 101 | |
| 7 | 1 | spilled |

**Statemachine** (benchmark document, 258 nodes - 99.2% inline, 0.8% spilled):

| Tokens | Nodes | |
|---|---|---|
| 1 | 7 | |
| 3 | 249 | states, events, transitions |
| 6 | 1 | spilled |
| 8 | 1 | spilled |

Note: Adding those additional pointers doesn't seem to have a measurable impact on the heap at all.

internal/generator/parser_generator.go
msujew: It's about the `var current AstNode` that every parser rule keeps track of. A lot of rules don't need to allocate that, since they are just `X: A | B | C`, which already allocates internally (inside of `A`, `B` or `C`). Therefore, no need to allocate a new AST node inside of the `X` parser function.

ast.go
Lotes: And what is the big difference to the previous implementation?
Now there is a buffer wired into the ASTNode. Before it was just assigned to the array field, also with 4 elements capacity.

msujew: The big difference is the lack of an allocation. Before:
1. Allocate the node
2. Once we parse a token, allocate a new slice for it.

Now:
1. Allocate the node. The allocator automatically allocates a 32-byte slot for the token array in the same call.

This cuts the amount of allocations required during parsing in half. Since everything else is already so efficient, these allocations are really expensive.

Lotes (approved):
LGTM.

PR #144, opened 2026-08-11
Implement linking optimizations

msujew:
The linking phase is usually the most performance intensive phase. This is an attempt to improve its performance by reducing the amount of allocations performed during the linking.

The main improvement is concerned with the new `SymbolSliceContainer`. Adopters (or the generated `<Language>SymbolContainer`) can implement this to generate allocation free scopes.

Another major improvement is the reduction of per-reference allocations. Mainly the new context that we generate for linking. It's now a single context with a special `resolutionChain` attached to it that keeps track of cycles. Adds a `ForEachWithSetup` for this, which keeps one setup function per worker.

Then there are a few other improvements:
1. Reduce the amount of allocations in the parser by storing the closures for reference resolving functions
2. Store the `name` of a symbol on the symbol description directly. This reduces the need for interface dispatches (expensive).
3. Reduce the amount of closures required for traversing the AST.
4. Caches the available number of CPU cores. Calling `GOMAXPROCS` on every `ForEach` call had to acquire a lock first internally.

All of these changes together roughly **double** the performance of the workspace build.

### DIFF
--- a/ast.go
+++ b/ast.go
@@ -248,10 +248,14 @@ func (node *AstNodeBase) Resolve(path FragmentPath) (AstNode, error) {
 //
 // Note that this function will traverse the entire subtree, without short-circuiting.
 func traverseContent(node AstNode, fn func(AstNode)) {
-	node.ForEachNode(func(child AstNode, containerField unique.Handle[string], index int) {
+	// The visitor closure is created once and recurses via itself; recursing
+	// through traverseContent would heap allocate a closure per visited node.
+	var visit func(child AstNode, containerField unique.Handle[string], index int)
+	visit = func(child AstNode, containerField unique.Handle[string], index int) {
 		fn(child)
-		traverseContent(child, fn)
-	})
+		child.ForEachNode(visit)
+	}
+	node.ForEachNode(visit)
 }
 
 // AllNodes creates an iterator over the given node and all its descendant nodes.
@@ -386,20 +390,32 @@ func AssignContainers(doc *Document) {
 }
 
 func doAssignContainers(doc *Document, root AstNode, references *[]UntypedReference) {
-	root.SetDocument(doc)
-	root.ForEachNode(func(child AstNode, containerField unique.Handle[string], index int) {
+	// Both closures are created once per document and track the current parent
+	// node mutably; capturing the parent per recursion level instead would heap
+	// allocate two closures per AST node.
+	current := root
+	var visitNode func(child AstNode, containerField unique.Handle[string], index int)
+	var visitReference func(ur UntypedReference, containerField unique.Handle[string], index int)
+	visitNode = func(child AstNode, containerField unique.Handle[string], index int) {
 		child.SetDocument(doc)
-		child.SetContainer(root, containerField, index)
-		doAssignContainers(doc, child, references)
-	})
-	root.ForEachReference(func(ur UntypedReference, containerField unique.Handle[string], index int) {
+		child.SetContainer(current, containerField, index)
+		prev := current
+		current = child
+		child.ForEachNode(visitNode)
+		child.ForEachReference(visitReference)
+		current = prev
+	}
+	visitReference = func(ur UntypedReference, containerField unique.Handle[string], index int) {
 		*references = append(*references, ur)
 		unit := ur.Unit()
 		if stringNode, ok := unit.(CompositeNode); ok {
 			stringNode.SetDocument(doc)
-			stringNode.SetContainer(root, containerField, index)
+			stringNode.SetContainer(current, containerField, index)
 		}
-	})
+	}
+	root.SetDocument(doc)
+	root.ForEachNode(visitNode)
+	root.ForEachReference(visitReference)
 }
 
 // NamedNode represents an [AstNode] whose name is accessible as a string in the Name field.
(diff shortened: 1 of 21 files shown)


internal/generator/linker_generator.go
```diff
@@ -291,6 +291,28 @@ func generateSymbolContainers(context *LinkerGeneratorContext) codegen.Node {
 	node.AppendLine("}")
 	node.AppendLine()
 
+	node.AppendLine("func (sc *", name, "SymbolContainer) ForTypeSlice(t reflect.Type) ([]*core.SymbolDescription, bool) {")
+	node.Indent(func(n codegen.Node) {
+		if len(sortedTargets) > 0 {
+			n.AppendLine("switch t {")
+			for _, target := range sortedTargets {
+				subtypes := subtypesInTargets(target, sortedTargets, context.ifaceParents)
```
ssmifi: _Minor_: `subtypesInTargets` is calculated twice.

linking/scoping.go
```diff
@@ -41,17 +40,33 @@ func LocalScopeOfType[T core.AstNode](node core.AstNode, globalScope core.Scope)
...
+// outerOrEmpty is the shortcut for nodes without local symbols: no new scope is created.
+func outerOrEmpty(outer core.Scope) core.Scope {
+	if outer != nil {
+		return outer
+	}
+	return core.EmptyScope
+}
+
 // GetLocalSymbols retrieves the local symbols for the given AST node and type.
 func GetLocalSymbols[T core.AstNode](node core.AstNode) core.SymbolSeq {
```
ssmifi: _Question_: This isn't called anywhere now. Is it kept intentionally as API?

textdoc/file.go
```diff
@@ -36,7 +36,8 @@ func NewFile(uri lsp.DocumentURI, languageID string, version int32, content stri
 		uri:        uri,
 		languageID: languageID,
 		version:    version,
-		content:    []byte(content),
+		// Convert in-place from string to []byte without copying.
+		content: unsafe.Slice(unsafe.StringData(content), len(content)),
```
ssmifi: _Question_: I am wondering if this can become unsafe in future edits, e.g., by changes to `Overlay.applyChangeLocked`. Should/Can we add a test asserting that edits never mutate the original input?

ast.go
```diff
@@ -386,20 +390,32 @@ func AssignContainers(doc *Document) {
 }
 
 func doAssignContainers(doc *Document, root AstNode, references *[]UntypedReference) {
-	root.SetDocument(doc)
-	root.ForEachNode(func(child AstNode, containerField unique.Handle[string], index int) {
+	// Both closures are created once per document and track the current parent
```
ssmifi: _Suggestion_: Should we add a test that checks if the container is actually correct to prevent future regressions?

reference_test.go
```diff
@@ -0,0 +1,32 @@
...
+		return nil, nil
+	})
+	done := make(chan struct{})
+	go func() {
+		ref.Resolve(context.Background())
+		close(done)
+	}()
+	select {
+	case <-done:
+	case <-t.Context().Done():
```
ssmifi: _Remark_: `t.Context().Done()` could take some time. Maybe something like `context.WithTimeout(t.Context(), 10*time.Second)` would provide quicker feedback.

ssmifi (commented):
Thanks, looks good already. Just a few minor questions/remarks.

ast.go
msujew: How? I don't think we have the meta model knowledge about what elements can have which container types during runtime. That would probably also incur quite a heavy performance cost.

textdoc/file.go
msujew: This behavior is already documented:

https://github.com/TypeFox/fastbelt/blob/ed879e869013c62929eca37cd41e009cbba5498b/textdoc/overlay.go#L203-L207

linking/scoping.go
msujew: Yes, it still makes sense as public API 👍

textdoc/file.go
ssmifi: That is probably all we can do for adopters. Would it still make sense to add a test for overlays to check that it does not mutate the original string to prevent accidental regression?

ast.go
ssmifi: I was thinking about another project, where a bug in the scoping actually corrupted the parent container. So, I was more thinking about a general containment regression test that just walks the ast a second time after assignment and not a runtime test.

E.g., sth like 
```go
func TestAssignContainersConsistency(t *testing.T) {
	f := test.New(t, CreateServices())
	doc := f.Parse(lightSwitch)
	doc.AssertNoErrors()

	root := doc.Root()
	if root.Container() != nil {
		t.Errorf("root node: expected nil container, got %v", root.Container())
	}

	var visited int
	var walk func(parent fastbelt.AstNode)
	walk = func(parent fastbelt.AstNode) {
		parent.ForEachNode(func(child fastbelt.AstNode, field unique.Handle[string], index int) {
			visited++
			gotField, gotIndex := child.ContainmentData()
			if child.Container() != parent {
				t.Errorf("node %T: expected container %T, got %T", child, parent, child.Container())
			}
			if gotField != field {
				t.Errorf("node %T: expected containment field %q, got %q", child, field.Value(), gotField.Value())
			}
			if gotIndex != index {
				t.Errorf("node %T: expected containment index %d, got %d", child, index, gotIndex)
			}
			walk(child)
		})
	}
	walk(root)

	if visited == 0 {
		t.Fatal("fixture produced no child nodes, test would trivially pass")
	}
}
```
So, it maybe was more of a question if sth like this would be worth it.

ssmifi (commented):
Thanks. Looks good. Just following up on the open questions.

PR #147, opened 2026-08-26
Support per-document read lock

msujew:
This PR attempts to fix a similar problem as our implementation already does inside of Langium. Some services require only specific document states to be available. For example, a document symbol request can be immediately resolved, once the document has been succesfully parsed - without waiting for the build to finish, or even references to be linked.

This change adds two main features for this:
1. Adds a `ReadAt` method to the `Lock` interface, which allows to read documents at specific states.
2. Adds a `DocumentStateRequirements` interface, that LSP services can (optionally) implement to start reading at a specific workspace/document state. Any service that does not implement this interface falls back to `Read`.

Removes the `downgrade` functionality from the lock, which has been replaced with `ReadAt`. `Write` and `Read` are now fully exclusive.

Related to https://github.com/eclipse-langium/langium/pull/2214.

### DIFF
--- a/server/completion_contributor.go
+++ b/server/completion_contributor.go
@@ -97,6 +97,9 @@ func NewDefaultCompletionContributor() CompletionContributor {
 	return &DefaultCompletionContributor{}
 }
 
+// Ensure DefaultCompletionContributor implements CompletionContributor.
+var _ CompletionContributor = (*DefaultCompletionContributor)(nil)
+
 // CompletionForToken emits a default item only for keyword tokens that contains a letter or digit.
 // This prevents trivial punctuation tokens from appearing in the default completion list.
 func (*DefaultCompletionContributor) CompletionForToken(_ context.Context, tt *core.TokenType, _ int, _ ContributorContext, accept CompletionAcceptor) {
--- a/workspace/builder.go
+++ b/workspace/builder.go
@@ -20,14 +20,20 @@ import (
 // linking, reference indexing, and validation.
 type Builder interface {
 	// Build processes the provided documents through all build phases (parse, compute
-	// symbol table, link, validate). It should regularly check ctx for cancellation
-	// between phases. downgrade must be called by the implementation once phases 1 and 2
-	// (the write phase) are complete; this transitions the workspace lock to readable so
-	// that read requests can proceed while phase 3 (validation) runs.
-	Build(ctx context.Context, docs []*core.Document, downgrade func()) error
+	// symbol table, link, validate). It must be called under [Lock.Write] and should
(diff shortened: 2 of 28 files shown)


server/completion_contributor.go
```diff
@@ -97,6 +97,9 @@ func NewDefaultCompletionContributor() CompletionContributor {
 	return &DefaultCompletionContributor{}
 }
 
+// Ensure DefaultCompletionContributor implements CompletionContributor.
+var _ CompletionContributor = (*DefaultCompletionContributor)(nil)
```
Lotes, line 101: Can you explain this line?

Looks like Typescript for:

```
const _: CompletionContributor|null = (DefaultCompletionContributor|null)null;

//where "DefaultCompletionContributor implements CompletionContributor"
```

It would make no sense in Typescript.

But we are not in Typescript... is it meant as a inheritance check?
"Does the DefaultCompletionContributor implement CompletionContributor???"

Lotes: I guess it is there to remind you to implement the interface entirely and exactly???

workspace/builder.go
```diff
@@ -69,25 +87,28 @@ func (s *DefaultBuilder) Build(ctx context.Context, docs []*core.Document, downg
...
 	})
 
 	if err := ctx.Err(); err != nil {
 		return err
 	}
+	// Phase 1 barrier: every document in docs is now parsed with exports
+	// computed, and documents outside docs were already complete when the
+	// updater collected the build set. Report the workspace floor so
+	// workspace-wide ReadAt calls are admitted.
+	lock.StateChanged(core.DocStateParsed | core.DocStateExportedSymbols)
```
Lotes, line 111: The comment makes sense, but in my eyes the method call is not reflecting the action very optimal. It reads like a notification, but is actually a wait for all go routines, or?!
Mh is this really a barrier? How did you set it up... like how does it know how many threads need to wait here?
I think I am misunderstanding here something...

Lotes: `WaitUntilAllDocumentsAreAtStates...`
`WaitUntilAllReached....`

server/completion_contributor.go
msujew: Yes, it's an inheritance check. The compiler omits the line, but it ensures that the type check will tell us that the struct actually implements the interface.

workspace/builder.go
msujew: It's a notification sent out to all goroutines that are currently in the `ReadAt` lock to re-evaluate their wait condition. If they waited on `Parsed`, they can now continue with their work - otherwise they are paused again until they are woken up.

Issue #140, opened 2026-08-06
[test] Improve completion testing infrastructure


msujew:
Right now, the infrastructure for testing completions is fairly barebones. It should work similar to diagnostic testing. I.e. `ExpectCompletion(label)` with additional chaining like `.Has(itemLabel)` and `.HasKind(itemLabel, kind)`, etc.

PR #149, opened 2026-09-06
feat(test): add ExpectCompletion chainable assertion API

Larslllllll:
## Summary

Add `ExpectCompletion(label)` to the `Doc` fixture — mirrors `ExpectDiagnostic` but for LSP completion items.

### `test/doc_fixture_lsp.go`
- Added `CompletionExpectation` struct with chainable methods:
  - `Has(label string)` — asserts a completion item with the given label exists
  - `HasKind(label string, kind lsp.CompletionItemKind)` — asserts label exists and has the given kind
  - `HasCount(n int)` — asserts exactly n items are present
  - `WithLabelPrefix(prefix string)` — asserts all items start with the given prefix
- Added `ExpectCompletion(label string)` on `*Doc` — calls `CompletionItems(label)` and returns a `*CompletionExpectation`

### `test/expect_completion_test.go`
- New test file with 9 passing tests demonstrating the API
- Tests cover: basic keyword completions, kind filtering, cross-reference completions, FQN composites, member call chains, optional groups, label prefix filtering

## Motivation

Resolves issue #140: "Improve completion testing infrastructure". The existing `CompletionItems(label)` helper returns a raw `[]lsp.CompletionItem` and requires manual label iteration. `ExpectCompletion` provides a fluent, diagnostic-friendly assertion API:

```go
// Before
items := doc.CompletionItems("cursor")
if !hasLabel(items, "foo") { t.Errorf(...) }

// After  
doc.ExpectCompletion("cursor").
    Has("foo").
    HasKind("foo", lsp.ReferenceCompletion)
```

## Testing

All 9 new tests pass. Full project test suite (`go test ./...`) passes with no regressions.

### DIFF
--- a/test/doc_fixture_lsp.go
+++ b/test/doc_fixture_lsp.go
@@ -1,6 +1,9 @@
 package test
 
 import (
+	"strings"
+	"testing"
+
 	core "typefox.dev/fastbelt"
 	"typefox.dev/fastbelt/server"
 	"typefox.dev/fastbelt/textdoc"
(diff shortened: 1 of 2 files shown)


msujew:
Hey @Larslllllll,

thanks for spending the tokens to generate this PR. However, as the issue is labeled a "Good First Issue", we want people to use this contribution to familiarize themselves with the code base. If we were just out to resolve this issue, we would be faster to just do this ourselves. I'll this PR for that reason.

Issue #138, opened 2026-08-06
[Validation] Accept service container in validation functions


msujew:
Right now, the validation functions have no way to access the current service container. This might be useful for adopters, since there might be some stateful services with caches, etc. that help do some of the work more efficiently.

PR #151, opened 2026-09-10
Add service container to validate

ssmifi:
Suggestion for https://github.com/TypeFox/fastbelt/issues/138

In fact, I would opt for making this explicit. Wdyt?

### DIFF
--- a/validation.go
+++ b/validation.go
@@ -4,7 +4,11 @@
 
 package fastbelt
 
-import "context"
+import (
+	"context"
+
+	"typefox.dev/fastbelt/util/service"
+)
 
 // DiagnosticSeverity mirrors LSP DiagnosticSeverity values.
 type DiagnosticSeverity int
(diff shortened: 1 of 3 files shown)


validation.go
```diff
@@ -76,9 +80,10 @@ type ValidationAcceptor func(diagnostic *Diagnostic)
 // Validator can be implemented by AST node Impl structs to provide custom validation checks.
 type Validator interface {
 	// Validate performs validation on the receiver node.
+	// sc is the service container of the running language server.
 	// The level parameter identifies when validation runs (e.g. "on-type", "on-save").
 	// The accept callback is used to collect diagnostics.
-	Validate(ctx context.Context, level string, accept ValidationAcceptor)
+	Validate(ctx context.Context, sc *service.Container, level string, accept ValidationAcceptor)
```
msujew: Suggestion: I would go one level further. Can we put the `level` inside of the `ctx` with a special key identifying the current level? 99% of validations probably don't need the level anyway.

msujew (commented):
I like the new API. I think we can improve on that though a tiny bit.

validation.go
ssmifi: Removed `level` as discussed.

workspace/document_validator.go
msujew: Suggestion: I think we can completely remove the `level string` from the `DocumentValidator` interface.

workspace/document_validator.go
```diff
@@ -13,10 +13,9 @@ import (
 
 // DocumentValidator validates a document's AST and collects diagnostics.
 type DocumentValidator interface {
-	// Validate returns diagnostics for doc at the given validation level
-	// (for example "on-save"). It walks doc.Root and calls [core.Validator]
-	// methods on AST nodes that implement that interface.
-	Validate(ctx context.Context, doc *core.Document, level string) []*core.Diagnostic
+	// Validate returns diagnostics for doc. It walks doc.Root and calls
+	// [core.Validator] methods on AST nodes that implement that interface.
```
msujew: Unrelated to this PR, but while you're at it: This comment explains what the default implementation of the interface does - not what the interface method should do. Can you adjust this a bit before merging?

msujew (approved):
Thanks, looks good to me. Please fix the git conflicts, before you can merge 👍

Issue #141, opened 2026-08-06
[CLI/Extension] Support rendering railroad diagrams for grammars


msujew:
Langium already offers to render railroad diagrams in its vscode extension and its CLI generator. We should be able to do the same (via a CLI plugin?). See https://github.com/tabatkins/railroad-diagrams for inspiration.

PR #152, opened 2026-09-13
Add railway diagrams

ssmifi:
STILL WIP, first draft open for discussion, still needs some polish.

Adds railway diagrams to the CLI (https://github.com/TypeFox/fastbelt/issues/141) _and_ to the hover.
image

Currently, the railway diagrams port (from https://github.com/tabatkins/railroad-diagrams) lives inside an own railway package. We can discuss if this should be released as own go port project.


cmd/fastbelt/generate.go
```diff
@@ -165,10 +166,33 @@ func runGenerateCLI(opts generateOptions) error {
...
+func writeRailroadDiagrams(g grammarPkg.Grammar, packageName, outputPath string, writeFile func(name, path, content string) error) error {
+	railroadDir := filepath.Join(outputPath, "railroad")
+	if err := os.MkdirAll(railroadDir, 0755); err != nil {
+		return err
+	}
+	diagrams := generator.GenerateRailroadDiagrams(g)
+	names := make([]string, 0, len(diagrams))
+	for name, svg := range diagrams {
+		names = append(names, name)
+		if err := writeFile("railroad-"+name, filepath.Join(railroadDir, name+".svg"), svg); err != nil {
```
Lotes: Idea: We could put all SVGs into a Markdown file using data urls.

internal/grammarhover/provider.go
```diff
@@ -0,0 +1,93 @@
...
+	rule, ok := node.(grammar.AbstractRule)
+	if !ok {
+		return ""
+	}
+	diagram, ok := railroad.BuildRuleDiagram(ctx, rule)
+	if !ok {
+		return ""
+	}
+	encoded := base64.StdEncoding.EncodeToString([]byte(diagram.SVG()))
+	return fmt.Sprintf("![Railroad diagram](data:image/svg+xml;base64,%s)", encoded)
```
Lotes: Magic 🌈!

internal/grammarhover/provider.go
```diff
@@ -0,0 +1,93 @@
...
+// use for it, so it doesn't belong in the generic framework package.
+// Second, it couldn't live in server even if that were desirable:
+// internal/railroad already imports internal/grammar (to map grammar AST
+// nodes to diagram nodes), and internal/grammar's own test files import
+// package server (to exercise the real language server) - so server
+// importing internal/grammar or internal/railroad, directly or transitively,
+// would close an import cycle for those tests. This package is the one
+// place allowed to depend on all three, since it's imported only by the
+// grammar language server's entrypoint (internal/grammar/server) and by
+// this package's own tests.
```
msujew: I don't think this is a good way to structure the code. I'm fine with keeping the railroad code in `internal/railroad`, but instead of using `internal/grammar` inside of `internal/railroad`, the relationship should be reversed. The grammar package should use `internal/railroad` to generate the svg. This immediately fixes the import cycle and removes the need for the `grammarhover` package.

msujew (commented):
I'm alright with having an `internal/railroad` package, but we have to improve the code organization for this. See comment below.

cmd/fastbelt/generate.go
Lotes: On the other side, the markdown gets very polluted with SVG code then...
I thought it could be better for the user opening just one rendered markdown. Similar to the ATN diagrams which are based on mermaid JS.

ssmifi: Both true. Have to think about this. Maybe something we don't have to decide imeediately.

